PR #27, opened 2020-02-19
removal and addition of constraints

zidanessf:
1. Three macros added in order to remove/add constraints. These operations are achieved by the following step:
**add constraints**:
step 1. Analyze the input expression (e.g. sum(model[:x][t] for t in 1:10) == 10), and replace the global/semi_local symbols/Exprs with their value. Return the "@constraint(.,.,.)" string.
step 2. Insert this string to model(code) using Regex.
**remove constraints**:
simply remove the "@constraint" string located by constraint reference using Regex
2. A comprehensive test of these macros is presented, including simple constraint removal/addition,
constraint removal/addition within for loops, and batch constraint removal/addition.

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -7,6 +7,7 @@ version = "0.1.0"
 Dualization = "191a621a-6537-11e9-281d-650236a99e60"
 JuMP = "4076af6c-e467-56ae-b986-b466b2749572"
 MathOptInterface = "b8f27783-ece8-5eb3-8dc8-9495eed66fee"
+Revise = "295af30f-e4ad-537b-8983-00126c2a3abe"
 
 [compat]
 Dualization = "~0.2.1"
--- a/src/bilevelJuMPExtension.jl
+++ b/src/bilevelJuMPExtension.jl
@@ -1,0 +1,170 @@
+macro delete_cons(model,cref)
+    model_string = string(model)
+    cref_string = string(cref)
+    s = quote
+        local con = $cref_string
+        # println(con)
+        local model_local = Meta.parse($model_string)
+        local semi_locals = Base.@locals # get the local vars within the loop
+        local model_fixed = remove_local_var(model_local,semi_locals)
+        local s = repr(eval(model_fixed))
+        local p = Regex("(?<=\\n)(.*\\bconstraint\\b.*\\b$con\\b.*)(?=\\n)")
+        # p = Regex("$con")
+        # println(p)
+        local snew = replace(s,p => "")
+        local model_new = Meta.parse(snew)#默认会套上一层quote "x" -> "quote x end"
+        eval(model_new)
+    end
+    return s
+end
+macro add_upper_cons(model,expression...)#Almost the same as add_lower_cons except for Upper()/Lower()
+    model_string = string(model)
+    expression_string = [string(e) for e in expression]
+    return quote
+    local model_local = Meta.parse($model_string)
+    local expression_local = [Meta.parse(x) for x in $expression_string]
+    @assert length($expression_string) <= 2
+    @assert length($expression_string) >= 1
+    local semi_locals = Base.@locals # get the local vars within the loop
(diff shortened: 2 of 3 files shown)


joaquimg:
Hi! Sorry for the delay (back from hollidays now).
Thanks for the interest!
Are you familiar with JuMP version 0.21.x?
In order to keep the (future) documentation as simples as possible, I would like to keep methods similar to JuMP itself. You can note that JuMP does not use macros to delete stuff. Is is possible to use that style?
Also, when naming User facing methods its good to follow the JuMP style guidelines. Actually it it good to follow it wherever you can.
About the “add” macros, we can go for ‘@lower_constraint’ and ‘@upper_constraint’.

Project.toml
```diff
@@ -7,6 +7,7 @@ version = "0.1.0"
 Dualization = "191a621a-6537-11e9-281d-650236a99e60"
 JuMP = "4076af6c-e467-56ae-b986-b466b2749572"
 MathOptInterface = "b8f27783-ece8-5eb3-8dc8-9495eed66fee"
+Revise = "295af30f-e4ad-537b-8983-00126c2a3abe"
```
joaquimg: We should not have revise as a dependency

src/bilevelJuMPExtension.jl
```diff
@@ -0,0 +1,170 @@
+macro delete_cons(model,cref)
```
joaquimg: Lets follow jump style and have delete as a function.

src/bilevelJuMPExtension.jl
```diff
@@ -0,0 +1,170 @@
...
+        local p = Regex("(?<=\\n)(.*\\bconstraint\\b.*\\b$con\\b.*)(?=\\n)")
+        # p = Regex("$con")
+        # println(p)
+        local snew = replace(s,p => "")
+        local model_new = Meta.parse(snew)#默认会套上一层quote "x" -> "quote x end"
+        eval(model_new)
+    end
+    return s
+end
+macro add_upper_cons(model,expression...)#Almost the same as add_lower_cons except for Upper()/Lower()
```
joaquimg: upper_constraint would be better.

src/bilevelJuMPExtension.jl
```diff
@@ -0,0 +1,170 @@
...
+        local constraint = "@constraint(Lower($blm),$(cref),$(ctr))"
+    end
+    local m = replace(s,p=>(pstr->"$constraint\n"*pstr))
+    local model_new = Meta.parse(m)
+    # @info("constraint ***$constraint*** added")
+    eval(model_new)
+    # println(m)
+    end
+end
+function remain_opt_var(model,expression,varlist,semi_locals,local_var=[])
```
joaquimg: It would be good to add docstrings to the new functions.
Also comments in macros are ver important, they tend to be obscure to most users.

test/manipulation_test.jl
```diff
@@ -0,0 +1,93 @@
+using Revise
+using JuMP,Gurobi,BilevelJuMP
```
joaquimg: Can you test with an open source one?
Also, lets keep each test inside a function.
And use the testing functions instead of just prints.
Also, need to integrate to runtests file.

test/manipulation_test.jl
```diff
@@ -0,0 +1,93 @@
+using Revise
```
joaquimg: We shouldnt need revise in a test file.

joaquimg:
bump

zidanessf:
> bump

sorry for delay(back from COV-2019 xD). My solution is simply manipulating the code that describes the bilevel problem. It is not very good since the whole problem needs to be rebuild after every modification of constraints.

joaquimg:
Hope you are feeling better.
Let me know if you want to proceed with this PR, if so you can start by looking at the suggestions I left.

joaquimg:
Feel free to reopen, I will close because its stalling

Issue #24, opened 2020-01-03
Query duals


joaquimg:
- [x] from lower level is easier in most KKT solve methods.
- [x] from upper level could be done when solving problems as NLPs.
- [x] and politely block getting duals from the upper level

joaquimg:
Add shape check to output proper shapes on duals of constraints

joaquimg:
Add proper dual status info.

PR #48, opened 2020-04-13
Query lower level duals

guilhermebodin:
related to #24


guilhermebodin:
Ipopt is erroring with bizarre values
Captura de Tela 2020-04-13 às 17 59 19

PR #59, opened 2020-06-11
add kwarg force::Bool=false to JuMP.fix method

NLaws:
This way the `force` option can be used with BilevelVariableRef's. See [JuMP.jl/variables.jl](https://github.com/jump-dev/JuMP.jl/blob/db80726ed15f72450afcf832eec9c89a1c1bff77/src/variables.jl#L540) for what `force` does.


joaquimg:
Nice, it is probably breaking due to the Ipopt update, I will fix tolerances.

PR #66, opened 2020-06-28
JuMP.solver_name expects JuMP.Model

NLaws:
Not 100% sure this is how it should be implemented, but it fixed the bug for me locally. Here is JuMP.solver_name:

https://github.com/jump-dev/JuMP.jl/blob/26c327931fec98fc0f558b89e2b7ce63452c93e6/src/JuMP.jl#L336

### DIFF
--- a/src/jump.jl
+++ b/src/jump.jl
@@ -764,7 +764,7 @@ function JuMP.show_backend_summary(io::IO, model::BilevelModel)
     if model.solver === nothing
         print(io, "No solver attached")
     else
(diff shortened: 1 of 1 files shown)


joaquimg:
Thanks!
There are many outcomes that might lead to some issue.
Do you have a minimal example so that I can add the proper fix?

NLaws:
@joaquimg Sorry I haven't been able to find the time to provide an example but looks like you fixed it!

PR #76, opened 2020-07-29
Many many fixes and benchmark files

### LINKED ISSUE #70: support value(::GenericAffExpr{Float64, BilevelJuMP.BilevelVariableRef})
When trying to get the value of a variable with type `GenericAffExpr{Float64,BilevelJuMP.BilevelVariableRef}` I get the following:
```
ERROR: MethodError: convert(::Type{Union{}}, ::Float64) is ambiguous. Candidates:
  convert(::Type{Union{}}, x) in Base at essentials.jl:169
  convert(::Type{T}, x::Number) where T<:Number in Base at number.jl:7
  convert(::Type{T}, arg) where T<:VecElement in Base at baseext.jl:8
  convert(::Type{T}, x::Number) where T<:AbstractChar in Base at char.jl:179
Possible fix, define
  convert(::Type{Union{}}, ::Number)
Stacktrace:
 [1] value(::GenericAffExpr{Float64,BilevelJuMP.BilevelVariableRef}, ::JuMP.var"#35#36"{Int64}) at /Users/nlaws/.julia/packages/JuMP/YXK4e/src/aff_expr.jl:157
 [2] value(::GenericAffExpr{Float64,BilevelJuMP.BilevelVariableRef}; result::Int64) at /Users/nlaws/.julia/packages/JuMP/YXK4e/src/aff_expr.jl:345
 [3] value(::GenericAffExpr{Float64,BilevelJuMP.BilevelVariableRef}) at /Users/nlaws/.julia/packages/JuMP/YXK4e/src/aff_expr.jl:345
```

I'm guessing that this could be fixed with a definition similar to the existing:
```
function JuMP.value(v::BilevelVariableRef)::Float64
    m = owner_model(v)
    solver = m.solver
    ref = solver_ref(v)
    return MOI.get(solver, MOI.VariablePrimal(), ref)
end
```
in BilevelJuMP/src/jump.jl

### LINKED ISSUE #71: Set solver attribute
I thank the developers for the efforts put in the new release.
I kindly ask them to implement the "set_optimizer_attribute" function or a similar method (I see it is commented in the source code, so I believe it is an open issue).
Unlikely without that feature I cannot run my code.
Currently I am trying to make it work with set_mode and set_optimizer functions; as I find a workaround I write here.

I solved with this workaround:
model = BilevelModel()
model.mode = BilevelJuMP.SOS1Mode()
set_optimizer(model, () -> Gurobi.Optimizer(...{parameters}...))

I don't know why but "set_mode" wasn't recognized as a function, but maybe it is because of the updates, however model.mode = ... is equivalent to set_mode (in the current version)

I don't know why but the workaround doesn't work in debug mode

### LINKED ISSUE #72: Error using value with expressions (appeared with new release)
Dear developers,

In the use of the new version I discovered that the value function does not work anymore with expressions, like the ones shown in the following code.
In the version 0.1.0 of BilevelJuMP they were working perfectly, now that I updated my repository it doesn't work anymore (it may be related also to the upgrade of JuMP). I hope it should be easy and fast to solve.

[error_source.txt](https://github.com/joaquimg/BilevelJuMP.jl/files/4973418/error_source.txt)

I realized that this is the same issue of #70 ; the attached source code can be a detailed example

The problem is caused mainly  at line 155 of file aff_expr of JuMP because the returned value of function promote_op becomes "union" instead of Float64. I crosschecked it with a similar example developed in JuMP only.

### LINKED ISSUE #73: MathOptInterface.Bridges.LazyBridgeOptimizer{Ipopt.Optimizer} does not support copy with names
Dear Developers,

The use of the Ipopt model with the new version and updated Ipopt returns the following error.
The attached file is an example of problem I am experiencing. I am not sure if this can be easily solved by setting "copy_names=false" in file jump line 98
, but I am not sure about the consequences.
[error_code.txt](https://github.com/joaquimg/BilevelJuMP.jl/files/4989675/error_code.txt)

### LINKED ISSUE #74: FortunyAmatMcCarlMode : primal/dual_big_M not implemented
I realized that although the parameters can be set, they are not used in the code and in fact the following code returns the following error:
KeyError: key MathOptInterface.VariableIndex(8) not found

The error occurs in function get_bounds called from add_complement.

```
using JuMP, BilevelJuMP, Gurobi

m = BilevelModel(Gurobi.Optimizer, mode = BilevelJuMP.FortunyAmatMcCarlMode(primal_big_M = 1000, dual_big_M = 1000))

@variable(Upper(m), 0 <= x <= 3 )
@variable(Lower(m), 0 <= y[1:2])
@variable(Lower(m), 0 <= z[1:2])

a = JuMP.Containers.DenseAxisArray([y[i]+ z[i] for i = 1:2], 1:2)
d = JuMP.Containers.SparseAxisArray(Dict((1, i) => y[i] for i in 1:2))

c = y[1] + y[2]

@constraint(Upper(m), x >=1)

@constraint(Lower(m), y[1] >= 0.5)
@constraint(Lower(m), y[2] >= 0.5)

@objective(Upper(m), Min, x)

@objective(Lower(m), Min, y[1]+z[2])

optimize!(m)


_a = value.(a)
_c = value.(c)
_d = value.(d)
```

joaquimg:
closes #70 
closes #71
closes #72
closes #73
closes #74


joaquimg:
@davide-f
You can test this branch.

davide-f:
Dear Joaquim,

Thank you very much, this branch is working for the moment. I will keep you
updated.
Eventually, do you have the plan for using the start values to set an
initial hint for the linking constraints?
I think it would be very useful. I am experiencing significant
computational problems and unlikely, despite being feasible, the solver
cannot find a feasible starting point due to the large number of binary/SOS
constraints, which most of them could be set given the value of the
constraints in the primal problem.
Setting starting points for dual variables is challenging, however, it may
be easier to set the starting values for the binary/SOS constraints.

What do you think? May I be of any help?

All the best,
Davide

PR #86, opened 2020-08-08
Automatic format print_lp

davide-f:
This formulation enables the automatic selection of the file type when the optimization model is saved.


joaquimg:
we just need tests to make sure it remains working

PR #99, opened 2021-02-11
Update to GitHub CI

guilhermebodin:


guilhermebodin:
Closes #98

joaquimg:
Awesome!
Just need to update the label on readme.
I would mark x86 as fails allowed. Is it possible?
Cbc and Ipopt might not be super consistent...

guilhermebodin:
I read something about `continue-on-error` but many people said it was catchy, maybe the best thing, for now, is just to remove x86

joaquimg:
fair enough!

PR #119, opened 2021-06-14
Introduction to theory of bilevel problems

HesamShaelaie:


docs/src/theory.md
```diff
@@ -0,0 +1,314 @@
...
+
+
+Example (1) indicates that bilevel problems are intrinsically nonconvex due to their hierarchical structure. It has been proven that LP-LP problems are strongly NP-hard; even checking local optimality of LP-LP problems is NP-hard. 
+
+
+##  Linear and Convex Follower Problems
+
+In this section, we take a closer look at bilevel problems with linear or convex followers. This is the class of bilevel problems that we mainly consider in this research. In case of LP-LP problems, we have the following property.
+
+!!!note Theorem 1. If the set of optimal solutions of an LP-LP problem is nonempty, then it contains at least one vertex of the shared constraint set Ω.
```
odow, line 161: These need to be 
```
!!! note "Theorem 1"
    If the set ...
```

docs/src/theory.md
```diff
@@ -0,0 +1,314 @@
...
+
+<div class="panel panel-warning">
+**Theorem 1.**
+{: .panel-heading}
+<div class="panel-body">
+
+If the set of optimal solutions of an LP-LP problem is nonempty, then it contains at least one vertex of the shared constraint set ``\Omega``.
+
+</div>
+</div>
```
odow, line 171: Guess this `<div>` stuff is just a copy-paste issue?

docs/src/theory.md
```diff
@@ -0,0 +1,314 @@
...
+
+<a id="31">[31]</a> 
+Candler, Wilfred, and Roger Norton. Multi-level programming and development policy. The World Bank, 1977.
+
+<a id="32">[32]</a> 
+DeNegre, Scott T., and Ted K. Ralphs. "A branch-and-cut algorithm for integer bilevel linear programs." Operations research and cyber-infrastructure. Springer, Boston, MA, 2009. 65-78.
+
+
+
+=======
```
odow, line 312: Remove, along with blank lines at end of file

docs/src/theory.md
```diff
@@ -0,0 +1,314 @@
...
+
+```math
+\begin{align}
+\mathcal{L}(x,y,\lambda)=f(y)+ \lambda^T g(x,y)
+\end{align}
+```
+is the Lagrangian function associated with the follower problem. Problem (BP-KKT) is equivalent to Problem (BP) in the following sense.
+
+<div class="panel panel-warning">
+**Theorem 2.**
```
odow, line 198: Same thing here

docs/src/theory.md
```diff
@@ -0,0 +1,314 @@
...
+& g(x,y) \leq 0,\\
+& y\in Y.\\
+	\end{align}
+```
+For linear bilevel problems (LBPs), both problems are linear problems (LPs). We thus call LBPs as LP-LP bilevel problems. Similarly, we consider, e.g., MILP-MILP problems, in which both problems (ILP) and (IFP) are mixed- integer linear problems (MILPs). Further, for the class of convex MIQP-QP problems, all involved constraints are assumed to be linear in x and y, and the two objective functions F and f are assumed to be convex-quadratic in x and y. This renders Problem (ILP) a convex mixed-integer quadratic problem (MIQP) and Problem (IFP) a convex quadratic problem (QP).
+
+
+## Computational Challenges
+
+Bilevel problems possess interesting properties and are computationally chal- lenging due to their hierarchical structure. This is demonstrated by the following example.
```
odow, line 127:
```suggestion
Bilevel problems possess interesting properties and are computationally challenging due to their hierarchical structure. This is demonstrated by the following example.
```

docs/src/theory.md
```diff
@@ -0,0 +1,314 @@
+
+# Theory of Bi-level Optimization
+## Introduction
+Standard single-level optimization problems consist of a single decision-maker who controls a set of decision variables to optimize its goal. However, in many real-world problems, several decision-makers with conflicting goals interact, and each individual decision impacts the decisions and outcomes of the others. Such interdependencies cannot be captured with the standard-setting of a single decision-maker, and other, more adequate, optimization problems are required [[1]](#1).
```
odow, line 4: I'm not sure we should copy verbatim the first paragraph of Kleinert's thesis, even if it is cited.

odow (commented):
Can you remove the `.DS_Store` files and add then to the `.gitignore` please?

We can discuss more in the call.

.gitignore
```diff
@@ -3,4 +3,7 @@
 *.jl.mem
 deps/deps.jl
 Manifest.toml
-docs/build
\ No newline at end of file
+docs/build
+.DS_Store
+*/.DS_Store
+*/*/.DS_Store
```
odow, line 9: There is also `docs/.DS_Store` below

docs/make.jl
```diff
@@ -9,7 +9,8 @@ makedocs(
     authors  = "Joaquim Garcia",
     pages   = [
         "Home" => "index.md",
-        "manual.md"
+        "manual.md",
+        "theory.md"
```
matbesancon, line 13:
```suggestion
        "theory.md",
```
overall, it's common in Julia to leave a trailing comma, that way when you add a new line you don't need to modify the last one

docs/src/theory.md
matbesancon: agreed, things should be much shorter, more informal in the documentation

PR #121, opened 2021-06-15
Add examples to documentation

odow:
@HesamShaelaie here's an example. You should make other files like `FOBP_example2.jl`. No need to touch the rest.

You can build the docs <s>by installing Julia 1.0</s> as follows:
```
cd BilevelJuMP
julia --project=docs
] instantiate
include("docs/make.jl")
```
Then you can open `docs/build/index.html` locally.

Once this PR is merged, let me know if you want a call to walk through this locally on your machine.

Here's the preview link:

https://joaquimg.github.io/BilevelJuMP.jl/previews/PR121/examples/FOBP_example2/

@joaquimg may need to add you as a collaborator to get the push-preview working. (@joaquimg, you should also protect the master branch to only accept PRs.)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -4,3 +4,4 @@
 deps/deps.jl
 Manifest.toml
 docs/build
+docs/src/examples/*.md
--- a/docs/Project.toml
+++ b/docs/Project.toml
@@ -2,6 +2,7 @@
 BilevelJuMP = "485130c0-026e-11ea-0f1a-6992cd14145c"
 Cbc = "9961bab8-2fa3-5c5a-9d89-47fab24efd76"
 Documenter = "e30172f5-a6a5-5a46-863b-614d45cd2de4"
+Ipopt = "b6b21f68-93f8-5de0-b562-5493be1d77c9"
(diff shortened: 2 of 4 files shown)


odow:
Huh. @joaquimg these tests passed?

odow:
@joaquimg I ran into that Cbc SOS bug so I switched to Ipopt. We should probably add a slack integer variable in `Cbc.optimize!` and then delete it after the solve?

joaquimg:
At some point Cbc started failing. That the biggest motivation for me switching to SCIP here.
What exactly do we need to do to solve Cbc issue with SOS1?
SHould we track in an issue?

odow:
It's this: https://github.com/coin-or/Cbc/issues/363

joaquimg:
I see... we need a "slack" binary in the SOS set now.
We could have an option in SOS1Mode to add that binary.
It's hacky but I don't see many other options. And nothing having Cbc is just bad.

odow:
I'll make a PR to Cbc to fix it.

PR #130, opened 2021-06-23
Example4

HesamShaelaie:


docs/src/examples/FOBP_example5.jl
```diff
@@ -0,0 +1,66 @@
...
+# Upper level variables
+@variable(Upper(model), y, start = 0)
+
+#Lower level variables
+@variable(Lower(model), x, start = 0)
+
+# Then we can add the objective and constraints of the upper problem:
+
+# Upper level objecive function
+@objective(Upper(model), Min, 0)
```
odow, line 44: @joaquimg:
```julia
  LoadError: The objective function `0` is not supported by JuMP.
  Stacktrace:
    [1] error(s::String)
      @ Base ./error.jl:33
    [2] set_objective_function(model::BilevelJuMP.UpperModel, func::Int64)
```

Perhaps this needs to be `0.0`?

odow, line 44:
```suggestion
@objective(Upper(model), Min, 0.0)
```

@HesamShaelaie Julia distinguishes between the integer `Int64` `0` and the floating point `Float64` `0.0`.

odow:
@joaquimg I think you need to take a look at this doctest failure. Are models with a constant objective expected to fail?

docs/src/examples/FOBP_example5.jl
```diff
@@ -0,0 +1,65 @@
...
+# Upper level variables
+@variable(Upper(model), y, start = 0)
+
+#Lower level variables
+@variable(Lower(model), x, start = 0)
+
+# Then we can add the objective and constraints of the upper problem:
+
+# Upper level objecive function
+@objective(Upper(model), Min, 0.0)
```
odow, line 44:
```suggestion
@objective(Upper(model), Min, 0 * y + 0)
```

I wonder if this will work

odow (approved):

PR #133, opened 2021-06-25
Update FOBP_example2.jl

HesamShaelaie:


odow (approved):

odow:
See how small PRs are a good idea? 

Here's the preview link:
https://joaquimg.github.io/BilevelJuMP.jl/previews/PR133/examples/FOBP_example2/

Whenever you make a PR, just replace "PR133" with that number. (It only works if you have `push -u joaquim`, and not a push to your branch.)

PR #136, opened 2021-06-26
example7

HesamShaelaie:


docs/src/examples/PHTP_example2.jl
```diff
@@ -0,0 +1,79 @@
...
+# Dempe.
+
+optimize!(model)
+primal_status(model)
+termination_status(model)
+
+# Auto testing
+@test objective_value(model) ≈ 0 atol=atol
+sol = vcat(value.(x), value.(y))
+@test sol ≈ [0 ; 0 ; -10; -10] || sol ≈ [0 ; 30; -10; 10] #atol=atol
```
odow, line 79: Two things

 1. Very pedantic, but: the red circle means that there is no blank line at the end of the file. It's standard practice to end each file with a blank line.
2. Can we add a comment about why the solution might be one of two solutions? Are they both optimal? Where only these two solutions?

docs/src/examples/PHTP_example2.jl
```diff
@@ -0,0 +1,79 @@
...
+
+# Lower objective function
+@objective(Lower(model), Min,(-x[1] + y[1] + 40)^2 + (-x[2] + y[2] + 20)^2)
+
+# Lower constraints
+@constraint(Lower(model),  [i=1:2],- x[i] + 2y[i] <= -10)
+@constraint(Lower(model), [i=1:2], y[i] >= -10)
+@constraint(Lower(model), [i=1:2], y[i] <= 20)
+
+# Initial Starting conditions  #src
```
joaquimg, line 67: remove?

docs/src/examples/PHTP_example2.jl
```diff
@@ -0,0 +1,79 @@
...
+# Upper level variables
+@variable(Upper(model), x[i=1:2], start = 0)
+
+# Lower level variables
+@variable(Lower(model), y[i=1:2], start = -10)
+
+# Then we can add the objective and constraints of the upper problem:
+
+# Upper level objecive function
+@objective(Upper(model), Min,2x[1] + 2x[2] -3y[1] - 3y[2] -60)
```
joaquimg, line 46: `@objective(Upper(model), Min, 2x[1] + 2x[2] -3y[1] - 3y[2] -60)`

docs/src/examples/PHTP_example2.jl
HesamShaelaie: how can I find out about the number of the optimal solutions?

odow (approved):

PR #137, opened 2021-06-27
example8

HesamShaelaie:


odow:
Looks like we might have to relax the tolerances even further in the other tests. @joaquimg do you often see this issue with Ipopt?

docs/src/examples/DTMP_example1.jl
```diff
@@ -0,0 +1,79 @@
+# # Foundations of Bilevel Programming: Example 8
+# This example is from the book Decomposition Techniques in Mathematical Programming
+# Chapter 7.2, page 281, [url](https://www.springer.com/gp/book/9783540276852)
+
+# Bold point: using upperonly and loweronly, using equality in constraints
+
+# Model of the problem
+# First level
+# ```math
+
```
joaquimg, line 10: remove space

docs/src/examples/DTMP_example1.jl
```diff
@@ -0,0 +1,79 @@
+# # Foundations of Bilevel Programming: Example 8
+# This example is from the book Decomposition Techniques in Mathematical Programming
+# Chapter 7.2, page 281, [url](https://www.springer.com/gp/book/9783540276852)
+
+# Bold point: using upperonly and loweronly, using equality in constraints
```
joaquimg, line 5: we should point out that: upper/loweronly creates variable only in the proper model and hence might generate slightly smaller reformulated models

docs/src/examples/DTMP_example1.jl
```diff
@@ -0,0 +1,79 @@
...
+
+using BilevelJuMP
+using Ipopt
+using JuMP
+using Test
+
+model = BilevelModel(Ipopt.Optimizer, mode = BilevelJuMP.ProductMode(1e-9))
+
+# Global variables
+atol = 1e-3
```
joaquimg, line 36: lets remove this and add it explicitly in the `@test` lines

odow (approved):

PR #138, opened 2021-06-27
example9

HesamShaelaie:


HesamShaelaie:
@odow, could you please check the model and tell me how I should write the model?
I have put time into it and come with something, but I am not sure if it is correct.

docs/src/examples/IEGT_example1.jl
```diff
@@ -0,0 +1,91 @@
...
+# \min 10y_1 + 12y_2 + 15y_3,\\
+# \notag s.t.\\
+# y_1 + y_2 + y_3 = 200,\\
+# y_1 \leq x,\\
+# y_2 \leq 150,\\
+# y_3 \leq 100,\\
+# y[i] \geq 0, \forall i \in I\\
+# ```
+
+# Model with doul variable
```
odow, line 25: dual

docs/src/examples/IEGT_example1.jl
```diff
@@ -0,0 +1,91 @@
+# # Foundations of Bilevel Programming: Example 9
+# This example is from the book Investment in Electricity Generation and Transmission. [url](https://www.springer.com/gp/book/9783319294995)
+
+# Bold point(s): using Dualof() in case we have information on the dual variable(s) of the lower level problem.
```
odow, line 4: I don't understand this line.

docs/src/examples/IEGT_example1.jl
```diff
@@ -0,0 +1,91 @@
...
+# y_1 + y_2 + y_3 = 200,\\
+# y_1 \leq x,\\
+# y_2 \leq 150,\\
+# y_3 \leq 100,\\
+# y[i] \geq 0, \forall i \in I\\
+# ```
+
+# Model with doul variable
+# ```math
+# \min 40000x + 8760*(10y_1-\lambda * y_1),\\
```
odow, line 27: Where does the 8760 come from? What is `\lambda`?

docs/src/examples/IEGT_example1.jl
```diff
@@ -0,0 +1,91 @@
+# # Foundations of Bilevel Programming: Example 9
+# This example is from the book Investment in Electricity Generation and Transmission. [url](https://www.springer.com/gp/book/9783319294995)
```
odow, line 2: Since this example is more realistic, it needs a description of the physical problem. What the variables represent, etc.

Some (not all. I can do better) of the JuMP tutorials do this:
https://jump.dev/JuMP.jl/stable/tutorials/Nonlinear%20programs/space_shuttle_reentry_trajectory/
https://jump.dev/JuMP.jl/stable/tutorials/Mixed-integer%20linear%20programs/sudoku/

docs/src/examples/IEGT_example1.jl
HesamShaelaie: I add the important points of the example. Better to say, what makes it different from other examples, and here is the usage of the Dualof() command in the model.

docs/src/examples/IEGT_example1.jl
HesamShaelaie: lambda is the dual variable of one of the second-level problem.

joaquimg: 8760 is the number of hours in the year.
that is part of the formulation in the paper. as @odow said we should add a tiny bit of context. Explaining physical meaning of variables and constraints....

docs/src/examples/IEGT_example1.jl
```diff
@@ -0,0 +1,91 @@
...
+# Second level
+# ```math
+# \min 10y_1 + 12y_2 + 15y_3,\\
+# \notag s.t.\\
+# y_1 + y_2 + y_3 = 200,\\
+# y_1 \leq x,\\
+# y_2 \leq 150,\\
+# y_3 \leq 100,\\
+# y[i] \geq 0, \forall i \in I\\
+# ```
```
joaquimg, line 23: we should go straight to model with dual

docs/src/examples/IEGT_example1.jl
```diff
@@ -0,0 +1,91 @@
...
+@constraint(Upper(model), ub0, x <= 250)
+
+# Followed by the constraints of the lower problem:
+@constraint(Lower(model), b, y[1] + y[2] + y[3] == 200)
+@constraint(Lower(model), ub1, y[1] <= x)
+@constraint(Lower(model), ub2, y[2] <= 150)
+@constraint(Lower(model), ub3, y[3] <= 100)
+@constraint(Lower(model), lb[i=1:3], y[i] >= 0)
+
+# Now defining the dual variables
```
joaquimg, line 74: highlight that `b` is a reference to a constraint. Making it explicit so that one can use it in the modeling.

docs/src/examples/IEGT_example1.jl
```diff
@@ -0,0 +1,91 @@
...
+# ```math
+# \min 40000x,\\
+# \notag s.t.\\
+# 0 \leq x \leq 250,\\
+# ```
+# Second level
+# ```math
+# \min 10y_1 + 12y_2 + 15y_3,\\
+# \notag s.t.\\
+# y_1 + y_2 + y_3 = 200,\\
```
odow, line 18:
```suggestion
# y_1 + y_2 + y_3 = 200, \quad [\lambda]\\
```

odow (commented):
This still needs a better intro explaining the problem, etc.

docs/src/examples/IEGT_example1.jl
joaquimg: Yes, we need a bit of story telling, nothing complex. something like:

In this example we consider a power system with two existing generators and one being planned.
Their costs are... and existing generators have... available capacity.
The problem is to optimize the size of the new generators given that....

PR #139, opened 2021-06-28
example10: second order cone

HesamShaelaie:


odow:
More doc failures. @joaquim is it really this common for Ipopt to be flakey? Can we do something about that?

HesamShaelaie:
@odow Could you please help me with this error?

odow:
So the error is:

> LoadError: MathOptInterface.UnsupportedConstraint{MathOptInterface.VectorOfVariables, MathOptInterface.SecondOrderCone}: `MathOptInterface.VectorOfVariables`-in-`MathOptInterface.SecondOrderCone` constraint is not supported by the model.

Which says that Ipopt does not support SecondOrderCone constraints. 

The solution is this:
https://github.com/joaquimg/BilevelJuMP.jl/blob/4b2839e6e00ede959494ad6998e389a92d2b0f81/test/solvers/ipopt.jl#L3-L6

wrap Ipopt in a bridge to convert the SOC constraint into the quadratic
```Julia
using JuMP, Ipopt
model = Model() do
    return MOI.Bridges.Constraint.SOCtoNonConvexQuad{Float64}(Ipopt.Optimizer())
end
@variable(model, x[1:3] >= 0)
@constraint(model, x in SecondOrderCone())
@objective(model, Min, x[1])
@constraint(model, sum(x[2:3]) >= 2)
optimize!(model)
```

odow (approved):

PR #141, opened 2021-07-04
valid inequalities

matbesancon:
Adding the first draft for valid inequalities.
There are still pieces missing I need to figure out and the main algorithm looping over constraints and tightening the bounds


matbesancon:
@joaquimg I'll need some help on some parts here, getting the variables from both levels and copying them into a new model that gets high point relaxation + dual feasibility constraints

matbesancon:
bump here

joaquimg:
this is in my todo list :)

Issue #124, opened 2021-06-19
Add MibS_jll as a backend


odow:
I present: calling MibS from Julia with automated binary installs on every platform!

cc @tkralphs

```Julia
julia> MibS_jll.mibs() do exe
           run(`$(exe) -Alps_instance model.mps -MibS_auxiliaryInfoFile model.aux`)
       end

========================================================
========================================================
Welcome to MibS (Mixed Integer Bilevel Solver)
Copyright (C) 2007-2019 Lehigh University, Scott DeNegre, Ted Ralphs
and Sahar Tahernejad.
All Rights Reserved.
This software is licensed under the Eclipse Public License. Please see
accompanying file for terms.
Version: 1.1.3
Build Date: Jun 18 2021
========================================================
========================================================
==  Welcome to the Abstract Library for Parallel Search (ALPS) 
==  Copyright 2000-2019 Lehigh University and others 
==  All Rights Reserved. 
==  Distributed under the Eclipse Public License 1.0 
==  Version: 1.5
==  Build Date: Jun 14 2021
Alps0030I Data file: model.mps
Coin0001I At line 1 NAME          F5010W01.KNP
Coin0001I At line 2 ROWS
Coin0001I At line 5 COLUMNS
Coin0001I At line 16 RHS
Coin0001I At line 18 BOUNDS
Coin0001I At line 29 ENDATA
Coin0002I Problem F5010W01.KNP has 1 rows, 10 columns and 10 elements
LL Data File: model.aux
Number of LL Variables:   10

=======================================
Analyzing problem structure            
=======================================

Number of UL Variables: 10
Number of LL Variables: 10
Number of UL Rows: 1
Number of LL Rows: 11
This instance is an interdiction problem.
This instance is a pure integer problem.
All upper level variables are binary.
All lower level variables are binary.
Coefficient matrix of upper level variables in upper level problem is non-negative.
Coefficient matrix of lower level variables in upper level problem is non-negative.
Coefficient matrix of upper level variables in lower level problem is non-negative.
Coefficient matrix of lower level variables in upper level problem is non-negative.

=======================================
Setting parameters                     
=======================================

Default branching strategy is MibSBranchingStrategyLinking.
Benders cut generator (just one cut) is on.
solveSecondLevelWhenXYVarsInt is set.
solveSecondLevelWhenLVarsFixed is set.
computeBestUBWhenLVarsFixed is set.
Linking solution pool will be used.

[...]

PR #142, opened 2021-07-04
initial start on in mps files

HesamShaelaie:


src/jump_input.jl
```diff
@@ -0,0 +1,67 @@
...
+)
+    # A new model to build
+    model = MOI.Utilities.Model{Float64}()
+    # Create a copy of the upper model
+    index_map = MOI.copy_to(model, upper)
+    # TODO ... add the lower problem to `model`
+    return model, upper_variables, lower_variables, lower_objective
+end
+
+
```
odow, line 50: Rather than testing the solution of the model, you should be testing the output of this function:
```julia
new_model, upper_variables, lower_variables, lower_objective = _build_single_model(model)
@test num_variables(new_model) == 2
@test length(upper_variables) == 1
@test length(lower_variables) == 1
@test objective_function(new_model) ≈ 3 * lower_variables[1] + upper_variables[1]
@test lower_objective ≈ -lower_variables[1]
# ... and so on
```

HesamShaelaie: Thanks, I was aware of this, I am playing around to learn what should I do.

src/jump_input.jl
```diff
@@ -0,0 +1,67 @@
...
+    lower::MOI.ModelLike, 
+    # A dictionary that maps variables in the upper to variables in the lower
+    link::Dict{MOI.VariableIndex,MOI.VariableIndex},
+)
+    # A new model to build
+    model = MOI.Utilities.Model{Float64}()
+    # Create a copy of the upper model
+    index_map = MOI.copy_to(model, upper)
+    # TODO ... add the lower problem to `model`
+    return model, upper_variables, lower_variables, lower_objective
```
odow, line 47: We also need the lower constraints

src/jump_input.jl
```diff
@@ -0,0 +1,139 @@
...
+@constraint(Lower(model), l2, 4x +  y >= 8)
+@constraint(Lower(model), l3, 2x +  y <= 13)
+@constraint(Lower(model), l4, 2x - 7y <= 0)
+
+function _build_single_model(
+    model::BilevelModel,
+)
+    upper = JuMP.backend(model.upper)
+    lower = JuMP.backend(model.lower)
+    linkUL = JuMP.index(model.upper_to_lower_link)
```
odow, line 28:
```suggestion
    upper_to_lower_link = JuMP.index(model.upper_to_lower_link)
```

src/jump_input.jl
```diff
@@ -0,0 +1,139 @@
...
+    linkLU = JuMP.index(model.lower_to_upper_link)
```
odow, line 29:
```suggestion
    lower_to_upper_link = JuMP.index(model.lower_to_upper_link)
```

src/jump_input.jl
```diff
@@ -0,0 +1,139 @@
...
+    upper::MOI.ModelLike, 
+    lower::MOI.ModelLike, 
+    # A dictionary that maps variables in the upper to variables in the lower
+    linkUL::Dict{MOI.VariableIndex,MOI.VariableIndex},
+    linkLU::Dict{MOI.VariableIndex,MOI.VariableIndex}
+)
+    # A new model to build
+    model = MOI.Utilities.Model{Float64}()
+    # Create a copy of the upper model
+    uppertomodel = MOI.copy_to(model, upper)
```
odow, line 44:
```suggestion
    upper_to_model_link = MOI.copy_to(model, upper)
```

src/jump_input.jl
```diff
@@ -0,0 +1,139 @@
...
+    upper_varibales = [index_map[k] for k in keys(linkUL)]
```
odow, line 45:
```suggestion
    upper_variables = [upper_to_model_link[k] for k in keys(upper_to_lower_link)]
```

src/jump_input.jl
```diff
@@ -0,0 +1,139 @@
...
+    # constraint of the lower level
+    # variable of the lower level
+    # variable of the upper level
+
+
+   # tp_primal_obj = MOI.get(lower, MOI.ObjectiveFunctionType()) 
+   # @assert tp_primal_obj !== nothing 
+    lower_primal_obj = MOI.get(lower, MOI.ObjectiveFunction{MOI.ScalarAffineFunction{Float64}}()) 
+    MOI.Utilities.map_indices(lower_primal_obj) do x
+        return uppertomodel[linkLU[x]]
```
odow, line 62:
```suggestion
        return upper_to_model_link[lower_to_upper_link[x]]
```

odow (commented):
The linking of variables between models is confusing. Maybe these names would make that clearer?

src/jump_input.jl
```diff
@@ -0,0 +1,139 @@
...
+    uppertomodel = MOI.copy_to(model, upper)
+    upper_varibales = [index_map[k] for k in keys(linkUL)]
+    lower_varibales = [index_map[k] for k in values(linkLU)]
+    # upper only variables 
+    
+    #print(index_map)
+
+    # objective function of the lower level
+
+    # constraint of the lower level
```
odow, line 53: Something like this should get you most of the way there for the lower level constraints
```Julia
lower_constraints = Any[]
for (F, S) in MOI.get(lower, MOI.ListOfConstraints())
    for ci in MOI.get(lower, MOI.ListOfConstraintIndices{F,S}())
        lower_f = MOI.get(lower, MOI.ConstraintFunction(), ci)
        set = MOI.get(lower, MOI.ConstraintSet(), ci)
        # TODO: map `lower_f` to `model_f`
        new_ci = MOI.add_constraint(model, model_f, set)
        push!(lower_constraints, new_ci)
    end
end
```

src/jump_input.jl
```diff
@@ -0,0 +1,139 @@
...
+
+function _build_single_model(
+    upper::MOI.ModelLike, 
+    lower::MOI.ModelLike, 
+    # A dictionary that maps variables in the upper to variables in the lower
+    linkUL::Dict{MOI.VariableIndex,MOI.VariableIndex},
+    linkLU::Dict{MOI.VariableIndex,MOI.VariableIndex}
+)
+    # A new model to build
+    model = MOI.Utilities.Model{Float64}()
```
odow, line 42:
```suggestion
    model = MOI.FileFormats.MPS.Model()
```

We may as well build the MPS model directly

joaquimg:
I will detail here, how we can obtain the set of variables in a model:

`BilevelModel.upper` variable can be split in three main sets:

1. Upper only variable (do not appear in lower)
2. Upper variables that are parameters to the lower model
3. Lower variables that are parameters to the upper model

Item 2 is the same as the set of keys in `BilevelModel.upper_to_lower_link`, i.e., `keys(model.upper_to_lower_link)`
Item 3 is the same as the set of values in `BilevelModel.lower_to_upper_link`, i.e., `values(model.lower_to_upper_link)`
Item 1 is what is left in upper.

Analogously,

`BilevelModel.lower` variable can be split in three main sets:

1. Lower only variable (do not appear in upper)
2. Lower variables that are parameters to the upper model
3. Upper variables that are parameters to the lower model

Item 2 is the same as the set of keys in `BilevelModel.lower_to_upper_link`, i.e., `keys(model.lower_to_upper_link)`
Item 3 is the same as the set of values in `BilevelModel.upper_to_lower_link`, i.e., `values(model.upper_to_lower_link)`
Item 1 is what is left in lower.

That being said, we should have additional fields: `upper_only` and `lower_only` that can be julia `Set`s

src/jump_input.jl
HesamShaelaie: great, thanks.

HesamShaelaie:
what is the difference between upper_to_lower_link and lower_to_upper_link variables? As far as I know, linking variables are a set of variables that appear in both upper and lower problems.

joaquimg:
`upper_to_lower_link` are decision variables in the upper problem, and fixed in the lower problem.
`lower_to_upper_link` are decision variables in the lower problem, and fixed in the upper problem.

joaquimg:
All the variables either belong to the upper model or to de lower model.
Some variables from one model might appear (as "constants") in the other model, these are the linking variables.
Consequently, linking variables might belong to the upper or to the lower model - never both.

joaquimg:
In the following description:

> BilevelModel.upper variable can be split into three main sets:

> 1 Upper only variable (do not appear in lower)
> 2 Upper variables that are parameters to the lower model
> 3 Lower variables that are parameters to the upper model

only 1 and 2 are decision variables of the upper model. 3 are som of the variables of the lower model that appear in the upper.

HesamShaelaie:
> In the following description:
> 
> > BilevelModel.upper variable can be split into three main sets:
> 
> > 1 Upper only variable (do not appear in lower)
> > 2 Upper variables that are parameters to the lower model
> > 3 Lower variables that are parameters to the upper model
> 
> only 1 and 2 are decision variables of the upper model. 3 are som of the variables of the lower model that appear in the upper.

got it thanks.

odow:
```Julia
model = _build(...)
MOI.write_to_file(model, "test.mps")
new_jump_model = JuMP.read_from_file("test.mps")
print(new_jump_model)
```

joaquimg:
After #146 , Master has `upper_only` and `lower_only` as julia `Set`s, can be used to find the variables that do not appear in both levels.
I also added comments in the main struct to facilitate development.

HesamShaelaie:
I was faced with this Err:

MethodError: no method matching MathOptInterface.NumberOfConstraints()

joaquimg:
For redirecting the STDOUT output of MibS:

https://github.com/jump-dev/AmplNLWriter.jl/blob/7d76863f08724f6f9f3bc5ce80e7afad36c134ac/src/AmplNLWriter.jl#L167

then you can parse the string

gurobi.log
```diff
@@ -0,0 +1,57 @@
...
+Cutting planes:
+  StrongCG: 1
+
+Explored 1 nodes (3 simplex iterations) in 0.01 seconds
+Thread count was 8 (of 8 available processors)
+
+Solution count 2: 6.57143 8 
+
+Optimal solution found (tolerance 1.00e-04)
+Best objective 6.571428571429e+00, best bound 6.571428571429e+00, gap 0.0000%
```
odow, line 57: Remove this file and add it to `.gitignore`

test/jump_input.jl
```diff
@@ -0,0 +1,183 @@
...
+# @test value(y) ≈ 8 / 15 atol=1e-6
+# @test value(u1) ≈ 3.5 * 8 / 15 atol=1e-6
+# @test value(l1) ≈ 4.5 * 8 / 15 atol=1e-6
+# @test dual(l1) ≈ [0] atol=1e-6
+# @test dual(l3) ≈ [0] atol=1e-6
+
+# TODO: why are these commented out?    #src
+# @test dual(l2) #≈ [0] atol=atol       #src
+# @test dual(l4) #≈ [0] atol=atol       #src
+=#
```
odow, line 183: We can remove all of these comments

src/test.mps
```diff
@@ -0,0 +1,43 @@
...
+    rhs       c5        0
+    rhs       u3        0
+    rhs       u4        0
+    rhs       c3        8
+RANGES
+BOUNDS
+ FR bounds    y
+ FR bounds    z
+ FR bounds    x
+ENDATA
```
odow, line 43: Remove this file

src/jump_input.jl
```diff
@@ -0,0 +1,135 @@
...
+)
+    upper = JuMP.backend(model.upper)
+    lower = JuMP.backend(model.lower)
+    
+    linkLU = Dict(
+        # lower to upper
+        JuMP.index(v) => JuMP.index(k) for (k,v) in model.link
+        # model.link means all link from upper to lower
+    )
+    linkLOnly = JuMP.index(model.lower_to_upper_link)
```
odow, line 14:
```suggestion
    linkLOnly = Dict(
        JuMP.index(k) => JuMP.index(v) for (k, v) in model.lower_to_upper_link
    )
```

odow (commented):
The tests are failing because the branch is out-dated. Here is how you can rebase onto the latest changes:
```
git checkout master 
git pull
git checkout MibSInput
git rebase master
git push -f
```

model.aux
```diff
@@ -0,0 +1,9 @@
+N	1
+M	4
+LC	2
+LR	3
+LR	8
+LR	4
+LR	5
```
odow, line 7: Something to check, but presumably these need to be ordered

HesamShaelaie:
@odow can we return the index of the integer constraint from the Bilevel model as well? This way, we can have Mixed mixed-integer input for the MibS. I know the first step is the run the linear-linear instances, but eventually, we need to go for mixed-mixed since MibS is tuned toward these kind of problems.

odow:
> @odow can we return the index of the integer constraint from the Bilevel model as well? This way, we can have Mixed mixed-integer input for the MibS. I know the first step is the run the linear-linear instances, but eventually, we need to go for mixed-mixed since MibS is tuned toward these kind of problems.

Do we need to tell the aux file that variables are integer? Aren't they just implied in the MPS file?

odow (commented):
So one way you could check if your aux files are correct is to pass them to the MibS executable and see if it finds the right result

model.mps
```diff
@@ -0,0 +1,43 @@
...
+	rhs       u3        0
+	rhs       u4        0
+	rhs       c3        8
+RANGES
+BOUNDS
+	FR bounds    y
+	FR bounds    z
+	FR bounds    x
+ENDATA
+	;
```
odow, line 43: Not sure what this `;` is doing there?

src/jump_input.jl
```diff
@@ -0,0 +1,198 @@
...
+    model::BilevelModel,
+)
+
+    new_model, lower_variables, lower_objective, lower_constraints, lower_sense  = _build_single_model(model)
+
+    # Write the MPS
+    MOI.write_to_file(new_model, "/Users/hesamshaelaie/Documents/BilevelJuMP.jl/src/model.mps")
+    
+    # Write the AUX
+    io = open("test.txt", "w");
```
odow, line 144: Move this to the function? also below. 

This should just be 
```Julia
write_auxillary_file(new_model, lower_variables, lower_objective, lower_constraints, lower_sense)
```

src/jump_input.jl
```diff
@@ -0,0 +1,198 @@
...
+    println(io, length(lower_variables))
+    println(io, length(lower_constraints))
+
+
+    
+    close(io);
```
odow, line 150: Instead of 
```Julia
io = open("test.aux", "w")
# ... stuff here
close(io)
```
prefer
```Julia
open("test.aux", "w") do io
    # ... stuff here
end
```

src/jump_input.jl
```diff
@@ -0,0 +1,198 @@
...
+    end
+
+    return the solution
+    =# 
+
+end
+
+
+
+function testing_solve_MibS()
```
odow, line 170: This should go in the `test` folder with the other tests

src/jump_input.jl
HesamShaelaie: got it.

src/jump_input.jl
HesamShaelaie: I will put it there eventually. It is here just for testing write_auxillary_file function, not for all the test sets.

model.mps
HesamShaelaie: what can I do about it?

src/jump_input.jl
```diff
@@ -0,0 +1,206 @@
...
+    Col = index_to_column_link(new_model)
+
+    open("model.aux", "w") do io 
+        println(io, "N    $(length(lower_variables))")
+        println(io, "M    $(length(lower_constraints))")
+
+        for x in 1:length(lower_variables)
+            vi = lower_variables[x]
+            println(io, "LC    $(Col[vi])")
+        end
```
odow, line 122:
```Julia
        for x in lower_variables
             println(io, "LC    $(Col[x])")
         end
```

src/jump_input.jl
```diff
@@ -0,0 +1,206 @@
...
+            for y in 1:length(lower_objective.terms)
+
+                vj = lower_objective.terms[y].variable_index
+
+                if vi == vj
+                    println(io, "LO    $(lower_objective.terms[y].coefficient)")
+                    break
+                end
+            end
+        end
```
odow, line 142: This doesn't account for variables missing objective coefficients.

```julia
obj_coefficients = Dict{MOI.VariableIndex,Float64}(
    x => 0.0 for x in lower_variables
)
for term in lower_objective.terms
    obj_coefficients[term.variable] += term.coefficient
end
for x in lower_variables
    println(io, "LO    $(obj_coefficients[x])")
end
```

src/jump_input.jl
```diff
@@ -0,0 +1,206 @@
...
+
+        
+        if lower_sense == MOI.MIN_SENSE
+            println(io, "OS    -1")
+        elseif lower_sense == MOI.MAX_SENSE
+            println(io, "OS    1")
+        else
+            # error
```
odow, line 150: The only other option is `MOI.FEASIBILITY_SENSE`, so just use `-1` instead. 
```julia
        println(io, "OS    ", lower_sense == MOI.MAX_SENSE ? 1 : -1)
```

odow:
Any progress on the "all linking variables should be discrete" error?

HesamShaelaie:
Hopefully, I will fix it tonight.  Could you please tell me how to get the result or log file out of each run of MibS. 

```julia
MibS_jll.mibs() do exe
        run(`$(exe) -Alps_instance model.mps -MibS_auxiliaryInfoFile model.aux`)
    end
```

odow:
https://github.com/joaquimg/BilevelJuMP.jl/issues/124#issuecomment-885868781

src/jump_input.jl
```diff
@@ -0,0 +1,298 @@
...
+    sleep(3)
+    MibS_jll.mibs() do exe
+        run(`$(exe) -Alps_instance $(add_mpx) -MibS_auxiliaryInfoFile $(add_aux)`)
+    end
+
+end
+
+
+
+function test_Writing_MibS_input_v1()
```
odow, line 187: Move to `tests`

knapsack.aux
```diff
@@ -0,0 +1,28 @@
+N 7
```
odow, line 1: Move all of these to `tests` if needed.

src/jump_input.jl
```diff
@@ -0,0 +1,298 @@
+using BilevelJuMP
+using JuMP
+using MibS_jll
+
+
```
odow, line 5: Formatting: don't have multiple empty lines

src/jump_input.jl
```diff
@@ -0,0 +1,298 @@
...
+function _build_single_model(
+    model::BilevelModel,
+)
```
odow, line 8: Formatting: `function _build_single_model(model::BilevelModel)`

src/jump_input.jl
```diff
@@ -0,0 +1,298 @@
...
+    upper = JuMP.backend(model.upper)
+    lower = JuMP.backend(model.lower)
+
```
odow, line 11: Formatting: remove unneeded empty lines within a function (here an everywhere else below)

src/jump_input.jl
```diff
@@ -0,0 +1,298 @@
...
+using MibS_jll
+
+
+function _build_single_model(
+    model::BilevelModel,
+)
+    upper = JuMP.backend(model.upper)
+    lower = JuMP.backend(model.lower)
+    
+    linkLU = Dict(
```
odow, line 12: Formatting: give a better name: link_all_lower_to_upper

src/jump_input.jl
```diff
@@ -0,0 +1,298 @@
...
+
+        # model.link means all link from upper to lower
+        JuMP.index(v) => JuMP.index(k) for (k,v) in model.link
+
+    )
+
+    linkLOnly = Dict(
```
odow, line 19: Formatting: give a better name: link_lower_only_to_upper

src/jump_input.jl
```diff
@@ -0,0 +1,298 @@
...
+    return _build_single_model(upper, lower, linkLU, linkLOnly)
+end 
+
+
+function _build_single_model(
+    upper::MOI.ModelLike, 
+    lower::MOI.ModelLike, 
+
+    # A dictionary that maps variables in the upper to variables in the lower
+    # upper_to_lower_link::Dict{MOI.VariableIndex,MOI.VariableIndex},
```
odow, line 31: Formatting: here and elsewhere: remove old comments

src/jump_input.jl
```diff
@@ -0,0 +1,298 @@
...
+
+    #upper_variables = [upper_to_model_link[k] for k in keys(upper_to_lower_link)]
+    lower_variables = [upper_to_model_link[k] for k in values(lower_only)]
+
+    
+    lower_constraints = Vector{MOI.ConstraintIndex}()
+    for (F, S) in MOI.get(lower, MOI.ListOfConstraints())
+        for ci in MOI.get(lower, MOI.ListOfConstraintIndices{F,S}())
+            
+                lower_f = MOI.get(lower, MOI.ConstraintFunction(), ci)
```
odow, line 51: Formatting: fix indentation

src/jump_input.jl
```diff
@@ -0,0 +1,298 @@
...
+    model = MOI.FileFormats.MPS.Model()
+    
+    # Create a copy of the upper model
+    upper_to_model_link = MOI.copy_to(model, upper)
+
+    #upper_variables = [upper_to_model_link[k] for k in keys(upper_to_lower_link)]
+    lower_variables = [upper_to_model_link[k] for k in values(lower_only)]
+
+    
+    lower_constraints = Vector{MOI.ConstraintIndex}()
```
odow, line 47:
```suggestion
    lower_constraints = MOI.ConstraintIndex[]
```

src/jump_input.jl
```diff
@@ -0,0 +1,298 @@
...
+    end
+
+    lower_sense = MOI.get(lower, MOI.ObjectiveSense())
+
+    return model, lower_variables, lower_objective, lower_constraints, lower_sense
+end
+
+
+function index_to_row_link(
+    model::MOI.FileFormats.MPS.Model)
```
odow, line 78: Formatting: `function index_to_row_link(model::MOI.FileFormats.MPS.Model)`

src/jump_input.jl
```diff
@@ -0,0 +1,298 @@
...
+            dict[ci] = i
+            i += 1
+        end
+    end
+    return dict
+end
+
+
+function index_to_column_link(
+    model::MOI.FileFormats.MPS.Model)
```
odow, line 96: Formatting: same comment as above

src/jump_input.jl
```diff
@@ -0,0 +1,298 @@
...
+end
+
+
+function write_auxillary_file(
+    new_model::MOI.FileFormats.MPS.Model,
+    lower_variables::Vector{MOI.VariableIndex},
+    lower_objective::MOI.ScalarAffineFunction,
+    lower_constraints::Vector{MOI.ConstraintIndex},
+    lower_sense::MOI.OptimizationSense,
+    aux_address::AbstractString
```
odow, line 111:
```suggestion
    aux_filename::String,
```

src/jump_input.jl
```diff
@@ -0,0 +1,298 @@
...
+    )
+
+    Row = index_to_row_link(new_model)
```
odow, line 114:
```suggestion
    row = index_to_row_link(new_model)
```
Variables should be lower-case

src/jump_input.jl
```diff
@@ -0,0 +1,298 @@
...
+    Col = index_to_column_link(new_model)
+
+    open(aux_address, "w") do io
```
odow, line 117:
```suggestion
    open(aux_filename, "w") do io 
```

src/jump_input.jl
```diff
@@ -0,0 +1,298 @@
...
+                obj_coefficients[term.variable_index] += term.coefficient
+            end
+        end
+        
+        for x in lower_variables
+            println(io, "LO $(obj_coefficients[x])")
+        end
+        
+        println(io, "OS ", lower_sense == MOI.MAX_SENSE ? -1 : 1)
+    end
```
odow, line 144:
```suggestion
    end
    return
```

src/jump_input.jl
```diff
@@ -0,0 +1,298 @@
...
+end
+
+
+
+function Writing_MibS_inputs(
```
odow, line 149:
```suggestion
function write_mibs_input_files(
```

src/jump_input.jl
```diff
@@ -0,0 +1,298 @@
...
+    model::BilevelModel,
+    name::AbstractString = "model"
```
odow, line 151:
```suggestion
    filename::String = "model"
```

src/jump_input.jl
```diff
@@ -0,0 +1,298 @@
...
+    name_mps = string(name, ".mps")
+    name_aux = string(name, ".aux")
+
+    MOI.write_to_file(new_model, name_mps)
+    write_auxillary_file(new_model, lower_variables, lower_objective, lower_constraints, lower_sense, name_aux)
+
+end
+
+function Running_MibS(
+add_mpx::AbstractString = "model.mps",
```
odow, line 167:
```suggestion
    mps_filename::String = "model.mps",
```

src/jump_input.jl
```diff
@@ -0,0 +1,298 @@
...
+    #name = "modelv2"
+    name_mps = string(name, ".mps")
+    name_aux = string(name, ".aux")
+
+    MOI.write_to_file(new_model, name_mps)
+    write_auxillary_file(new_model, lower_variables, lower_objective, lower_constraints, lower_sense, name_aux)
+
+end
+
+function Running_MibS(
```
odow, line 166:
```suggestion
function run_mibs_binary(
```

src/jump_input.jl
```diff
@@ -0,0 +1,298 @@
...
+add_mpx::AbstractString = "model.mps",
+add_aux::AbstractString = "model.aux"
```
odow, line 168:
```suggestion
    aux_filename::String = "model.aux"
```

src/jump_input.jl
```diff
@@ -0,0 +1,298 @@
...
+)
+    println("============================")
+    println("============================")
+    println("============================")
+    println(add_mpx)
+    println(add_aux)
+    println("============================")
+    println("============================")
+    println("============================")
+    sleep(3)
```
odow, line 178: Delete all of these

src/jump_input.jl
```diff
@@ -0,0 +1,298 @@
...
+    MibS_jll.mibs() do exe
+        run(`$(exe) -Alps_instance $(add_mpx) -MibS_auxiliaryInfoFile $(add_aux)`)
+    end
```
odow, line 181: Use the form I described in #124 with `pipeline` and the `IOBuffer`. We need to be able to parse the solution.

src/jump_input.jl
```diff
@@ -0,0 +1,421 @@
...
+                Dict_Lower_Name[CntD] = y
+                Dict_Lower_Value[y] = 0
+                CntD = CntD + 1
+            else
+                Dict_Upper_Name[CntU] = y
+                Dict_Upper_Value[y] = 0
+                CntU = CntU + 1
+            end
+        end
+
```
odow, line 260: Have you seen regex before? You should do something like this:

```Julia
function parse_output(output)
    lines = split(output, '\n')
    found_status = false
    objective_value = NaN
    upper = Dict{Int,Float64}()
    lower = Dict{Int,Float64}()
    for line in lines
        if !found_status
            if occursin("Optimal solution", line)
                found_status = true
            end
            continue
        end
        m = match(r"([xy])\[([0-9]+)\] \= (.+)", line)
        if m === nothing
            m = match(r"Cost \= (.+)", line)
            if m !== nothing
                objective_value = parse(Float64, m[1])
            end
            continue
        end
        column = parse(Int, m[2])
        value = parse(Float64, m[3])
        if m[1] == "x"
            upper[column] = value
        else
            lower[column] = value
        end
    end
    return objective_value, upper, lower
end
```

odow:
https://github.com/jump-dev/MathOptInterface.jl/blob/697ec4b93b12676fd01c066efde5b8d77cb0625f/src/Utilities/functions.jl#L6-L12

```
MOI.Utilities.eval_variables(objective) do variable
     # Variable is the MOI.VariableIndex. This function should return the value
    return # ...
end
```

src/jump_input.jl
```diff
@@ -1,4 +1,4 @@
-import MibS_jll
```
odow: You need MibS. CI is broken

Project.toml
```diff
@@ -10,6 +10,7 @@ JuMP = "4076af6c-e467-56ae-b986-b466b2749572"
 LinearAlgebra = "37e2e46d-f89d-539d-b4ee-838fcccc9c8e"
 MathOptInterface = "b8f27783-ece8-5eb3-8dc8-9495eed66fee"
 MibS_jll = "118347d2-e127-56b9-9933-6abf9cc1adc1"
+Revise = "295af30f-e4ad-537b-8983-00126c2a3abe"
```
odow, line 13: Remove from the Project.toml

Project.toml
```diff
@@ -9,6 +9,8 @@ IntervalArithmetic = "d1acc4aa-44c8-5952-acd4-ba5d80a2a253"
 JuMP = "4076af6c-e467-56ae-b986-b466b2749572"
 LinearAlgebra = "37e2e46d-f89d-539d-b4ee-838fcccc9c8e"
 MathOptInterface = "b8f27783-ece8-5eb3-8dc8-9495eed66fee"
+MibS_jll = "118347d2-e127-56b9-9933-6abf9cc1adc1"
+Revise = "295af30f-e4ad-537b-8983-00126c2a3abe"
```
matbesancon, line 13: the project should not depend on Revise, it's a local development tool for us

src/jump_input.jl
```diff
@@ -0,0 +1,223 @@
...
+        all_upper = Dict_Upper_Value,
+        all_lower = Dict_Lower_Value
+    )
+end
+
+# input
+# fails on ,...
+# outputs
+
+function solve_with_MibS(model::BilevelModel; silent::Bool = true)
```
odow, line 202:
```suggestion
"""

    solve_with_MibS(model::BilevelModel; silent::Bool = true)
    
Explain what the outputs are so that Joaquim can finish the rest.
"""
function solve_with_MibS(model::BilevelModel; silent::Bool = true)
```

src/jump_input.jl
```diff
@@ -0,0 +1,223 @@
...
+        if m[1] == "x"
+            upper[column] = value
+            nameofvar = Dict_Upper_Name[column]
+            Dict_Upper_Value[nameofvar] = value
+        else
+            lower[column] = value
+            nameofvar = Dict_Lower_Name[column]
+            Dict_Lower_Value[nameofvar] = value
+        end
+    end
```
odow, line 186: Instead of the names, we should return the maps in the space of the original variable indices. 

`x[0]` -> first upper variable in `new_model` -> variable in `BilevelModel.upper`

test/mibs.jl
```diff
@@ -0,0 +1,342 @@
...
+            end
+        end
+    end
+end
+
+function test_basic_example_1()
+    model = BilevelModel()
+    @variable(Upper(model), y, Int)
+    @variable(Upper(model), z, Int)
+    @variable(Lower(model), x, Int)
```
odow, line 21: TODO 1:  add an error message if not all variables are binary or integer to avoid MibS infinite loop.

src/jump_input.jl
odow: Example: https://github.com/odow/SDDP.jl/blob/3e14562b73db19b4267f3d42c113154093aa5b6e/src/algorithm.jl#L1269-L1285

odow:
Stretch goal: add another literate example calling MibS.

odow:
Another TODO: 

Try different simple problems to see which crash:

- MIP upper - LP lower
- continuous in upper
- etc

We should document which cases fail.

odow:
Final TODO: add one of the examples to your blog post.

src/jump_input.jl
odow: In the documentation outline the assumptions: i.e., that all variables are integer.

src/jump_input.jl
```diff
@@ -41,7 +42,16 @@ function _build_single_model(
     lower_objective = MOI.Utilities.map_indices(lower_objective) do x
         return upper_to_model_link[lower_to_upper_link[x]]
     end
-    #number varibles == inter + binary const
+
+    # Testing if the model is MIP-MIP or not. 
+    if check_MIPMIP
+        int_var = MOI.get(model, MOI.NumberOfConstraints{MOI.SingleVariable, MOI.Integer}())
```
odow, line 48: We should also check for `MOI.ZeroOne` variables. (I assume MibS can handle binary?)

Project.toml
odow, line 13:
```suggestion
```

src/jump_input.jl
```diff
@@ -0,0 +1,233 @@
+
```
odow, line 1:
```suggestion
using MibS_jll
```

src/jump_input.jl
```diff
@@ -209,6 +209,31 @@ end
 # fails on ,...
 # outputs
 
+"""
+    solve_with_MibS(
+                    model::BilevelModel; 
+                    silent::Bool = true)
```
odow, line 211: Please format as `solve_with_MibS(model::BilevelModel; silent::Bool = true)`

src/jump_input.jl
```diff
@@ -209,6 +209,31 @@ end
...
+
+## Inputs
+Inputs of the function are consist of one `model` with the data structure of MathOptInterface and `silent` which is an attribute for MibS solver to show the details of the solution or not (`true` value has been considered to hide the information).
```
odow, line 214:
```suggestion

 - `model::BilevelModel`:  the model. to optimize
 - `silent::Bool = true`: controls the verbosity of the solver. If `silent`,
   nothing is printed. Set to `false` to display the MibS output.
```

src/jump_input.jl
```diff
@@ -209,6 +209,31 @@ end
...
+"""
+    solve_with_MibS(
+                    model::BilevelModel; 
+                    silent::Bool = true)
+
+## Inputs
+Inputs of the function are consist of one `model` with the data structure of MathOptInterface and `silent` which is an attribute for MibS solver to show the details of the solution or not (`true` value has been considered to hide the information). 
+
+## Outputs
+For the output, we have multiple values, which are:
```
odow, line 217:
```suggestion

This function returns a `NamedTuple` with fields:
```

src/jump_input.jl
```diff
@@ -209,6 +209,31 @@ end
...
+* `status`: if the problem is feasible and has optimal value(s), it returns `true`.
```
odow, line 218:
```suggestion
* `status::Bool`: `true` if the problem is feasible and has an optimal solution. `false` otherwise.
```

src/jump_input.jl
```diff
@@ -209,6 +209,31 @@ end
...
+* `objective`: it is the objective value (cost) of the bilevel problem
+* `nonzero_upper`: it returns `dict{index => value}`, in which the `index` refers to the index of upper variables with non zero values and the index starts from `0`. Here the order of the variables is based on their order of appearance in the MPS file.
+* `nonzero_lower`: it has the same structure as `nonzero_upper`, but it represents the index of non-zero variables in the lower problem. 
+* `all_upper`: it returns all the upper variables, including zero and non-zero variables. For recalling the variables, you need to use the same name as you used to define the variables, e.g., for `@variable(Upper(model), y, Int)`, we need to use `all_upper["y"]` to get the value of the variable `y`.
+* `all_lower`: it has the same structure as the `all_upper` but is defined for lower variables.
+
+
+## Note
+Currently, `MibS` is designed to solve only MIP-MIP problems. Thus, if you define LP-MIP, MIP-LP, or LP-LP, it will throw an error.
```
odow, line 227:
```suggestion
!!! warning
    Currently, `MibS` is designed to solve MIP-MIP problems only. Thus, if
    you define LP-MIP, MIP-LP, or LP-LP, it will throw an error. 
```

src/jump_input.jl
```diff
@@ -209,6 +209,31 @@ end
...
+* `all_lower`: it has the same structure as the `all_upper` but is defined for lower variables.
+
+
+## Note
+Currently, `MibS` is designed to solve only MIP-MIP problems. Thus, if you define LP-MIP, MIP-LP, or LP-LP, it will throw an error. 
+
+"""
+
+
+
```
odow, line 232: Remove these blank lines. The `"""` must be immediately above the function

src/jump_input.jl
```diff
@@ -209,6 +209,31 @@ end
...
+* `objective`: it is the objective value (cost) of the bilevel problem
+* `nonzero_upper`: it returns `dict{index => value}`, in which the `index` refers to the index of upper variables with non zero values and the index starts from `0`. Here the order of the variables is based on their order of appearance in the MPS file.
+* `nonzero_lower`: it has the same structure as `nonzero_upper`, but it represents the index of non-zero variables in the lower problem. 
+* `all_upper`: it returns all the upper variables, including zero and non-zero variables. For recalling the variables, you need to use the same name as you used to define the variables, e.g., for `@variable(Upper(model), y, Int)`, we need to use `all_upper["y"]` to get the value of the variable `y`.
+* `all_lower`: it has the same structure as the `all_upper` but is defined for lower variables.
+
+
+## Note
+Currently, `MibS` is designed to solve only MIP-MIP problems. Thus, if you define LP-MIP, MIP-LP, or LP-LP, it will throw an error. 
+
```
odow, line 228: Remove blank line

src/jump_input.jl
```diff
@@ -209,6 +209,31 @@ end
...
+## Outputs
+For the output, we have multiple values, which are:
+* `status`: if the problem is feasible and has optimal value(s), it returns `true`.
+* `objective`: it is the objective value (cost) of the bilevel problem
+* `nonzero_upper`: it returns `dict{index => value}`, in which the `index` refers to the index of upper variables with non zero values and the index starts from `0`. Here the order of the variables is based on their order of appearance in the MPS file.
+* `nonzero_lower`: it has the same structure as `nonzero_upper`, but it represents the index of non-zero variables in the lower problem. 
+* `all_upper`: it returns all the upper variables, including zero and non-zero variables. For recalling the variables, you need to use the same name as you used to define the variables, e.g., for `@variable(Upper(model), y, Int)`, we need to use `all_upper["y"]` to get the value of the variable `y`.
+* `all_lower`: it has the same structure as the `all_upper` but is defined for lower variables.
```
odow, line 223: You get the idea with the rest of the fields. Give their type, and be more succinct.

src/jump_input.jl
```diff
@@ -225,7 +250,7 @@ function solve_with_MibS(model::BilevelModel; silent::Bool = true)
             aux_filename,
         )
         output = _call_mibs(mps_filename, aux_filename)
-        if silent
+        if !silent
```
odow, line 249: Ha! I think this was my fault.

src/jump_input.jl
```diff
@@ -206,30 +206,22 @@ function _parse_output(
...
+* `objective::Float64`: objective value (cost) of the problem
+* `nonzero_upper::Dict{Int, Float64}`: it returns `Dict{index => value}`, in which the `index` refers to the index of upper variables with non zero values and the index starts from `0`. Here, the order of the variables is based on their order of appearance in the MPS file.
+* `nonzero_lower::Dict{Int, Float64}`: it has the same structure as `nonzero_upper`, but it represents the index of non-zero variables in the lower problem. 
+* `all_upper::Dict{String, Float64}`: it returns `Dict{name => value}` which contains all upper variables values (zero and non-zero). For recalling the variables, you need to use the same name as you used to define the variables, e.g., for `@variable(Upper(model), y, Int)`, we need to use `all_upper["y"]` to get the value of the variable `y`.
+* `all_lower::Dict{String, Float64}`: it has the same structure as the `all_upper` but is defined for lower variables.
+!!! warning
+Currently, `MibS` is designed to solve MIP-MIP problems only. Thus, if you define LP-MIP, MIP-LP, or LP-LP, it will throw an error.
```
odow, line 223: The indentation here was crucial. It needs to be 
```
!!! warning
    Currently,  ...
```

This gets seen by Documenter and formatted appropriately.

There should also be a blank line between the `*` and the `!!!`

odow:
You can see how the docstring is formatted by typing `?` and then the name of the function into the REPL:

image

src/jump_input.jl
```diff
@@ -0,0 +1,246 @@
...
+        all_upper = Dict_Upper_Value,
+        all_lower = Dict_Lower_Value
+    )
+end
+
+"""
+    solve_with_MibS(model::BilevelModel; silent::Bool = true)
+
+## Inputs
+* `model::BilevelModel`:  the model. to optimize
```
odow, line 212:
```suggestion
* `model::BilevelModel`:  the model to optimize
```

src/jump_input.jl
```diff
@@ -0,0 +1,246 @@
...
+* `silent::Bool = true`: controls the verbosity of the solver. If `silent`, nothing is printed. Set to `false` to display the MibS output.
+## Outputs
+This function returns a `NamedTuple` with fields:
+* `status::Bool`: `true` if the problem is feasible and has an optimal solution. `false` otherwise.
+* `objective::Float64`: objective value (cost) of the problem
```
odow, line 217:
```suggestion
* `objective::Float64`: objective value (cost) of the upper problem
```

Did we do anything about the objective of the lower?

HesamShaelaie:
> You can see how the docstring is formatted by typing `?` and then the name of the function into the REPL:
> 
> image

HesamShaelaie:
Sorry for closing the pull request. It was a mistake.

src/jump_input.jl
HesamShaelaie: do you want me to test first if we can solve binary variables with our function and then go for it?

HesamShaelaie:
I updated the [GSoC](https://github.com/HesamShaelaie/HesamShaelaie.github.io/commit/3ed92439b2e73be281fb552572a958877c11961a) blog.

odow (approved):

src/jump_input.jl
odow: We can do this in a follow-up pr.

PR #150, opened 2021-08-15
Tidy up MibS PR

odow:
@HesamShaelaie this is a PR onto your branch. Take a look at the changes I made. I'll go through and annotate the particular bits.


test/mibs.jl
```diff
@@ -1,155 +1,63 @@
+module TestJuMPInput
```
odow, line 1: You can run the test file in the REPL as follows

```julia
include("test/mibs.jl")
```

If you're making changes to the package code, either restart the REPL every time, or use Revise.jl:
```Julia
using Revise
include("test/mibs.jl")
```

test/mibs.jl
```diff
@@ -1,155 +1,63 @@
...
+    @test MOI.get(new_model, MOI.NumberOfConstraints{MOI.ScalarAffineFunction{Float64},MOI.GreaterThan{Float64}}()) == (2 + 1)
+    @test MOI.get(new_model, MOI.NumberOfConstraints{MOI.SingleVariable, MOI.Integer}()) == 3
     x = lower_variables[1]
     @test lower_objective ≈ MOI.ScalarAffineFunction{Float64}([MOI.ScalarAffineTerm(-1.0, x)],0.0)
     @test lower_sense == MOI.MIN_SENSE
+    solution = BilevelJuMP.solve_with_MibS(model, silent = false)
+    @test solution.status == true
+    @test solution.objective ≈ 8
+    @test solution.nonzero_upper == Dict(0 => 8)
+    @test solution.nonzero_lower == Dict{Int,Float64}()
```
odow, line 48: For now we just return this solution struct. @joaquimg can handle mapping the solutions back into JuMP-space. That might be a little complicated.

src/jump_input.jl
```diff
@@ -1,445 +1,177 @@
...
-                
-                if startswith(line, "x")
-                    VarName = Dict_Upper_Name[VarIndex]
-                    Dict_Upper_Value[VarName] = Num
-                else
-                    VarName = Dict_Lower_Name[VarIndex]
-                    Dict_Lower_Value[VarName] = Num
-                end
-            end            
+        m = match(r"([xy])\[([0-9]+)\] \= (.+)", line)
```
odow, line 132: Regex are magical if you haven't seen them before, but complicated to understand.

`r"([xy])\[([0-9]+)\] \= (.+)"`

* Things  in `()` get "matched" so the first thing in `()` is `m[1]`, the second `m[2]` and so on.
* `([xy])` says "match the character x or the character y
* `\[` says match exactly `[` (you have to escape it because it's a special character
* `([0-9]+)` says match a sequence of numbers from 0-9 where there is at least one character. So `0`, `11`, `124` and so on
* `]` says match exactly `]`
* ` \= ` says match exactly ` = `
* `(.+)` says match everything else until the end of the line.

odow:
@HesamShaelaie do you have permission to merge this? If so, you might want to just click merge and then run `git pull` locally to see the changes.

odow:
I found the reason for the doc failure. It's my fault with the versions

odow:
Docs should work after https://github.com/JuliaRegistries/General/pull/42935 is merged. ~15 minutes

odow (commented):
So the bits still to do are:

- [ ] Add a docstring to the `solve_mibs` function
- [ ] Add something to the main documentation
- [ ] Add a check that the problem is MIP/MIP so it doesn't get caught in an infinite loop on solve

src/jump_input.jl
HesamShaelaie: it makes life soo easy.

PR #162, opened 2022-03-03
Add agg sum and rename set ub/lb to hint

joaquimg:


joaquimg:
@odow very weirdly MibS just do nothing the first time it's called on windows. Linux is fine.
https://github.com/joaquimg/BilevelJuMP.jl/runs/5418927087?check_suite_focus=true

joaquimg:
If I just send log to screen, things seem to work:
https://github.com/joaquimg/BilevelJuMP.jl/runs/5419284552?check_suite_focus=true

joaquimg:
I am writing something in the io before capturing mibs output.
Seems that misb is doing some delete(?) in the IO...

joaquimg:
aaand write to file then read the file works....
https://github.com/joaquimg/BilevelJuMP.jl/runs/5419572854?check_suite_focus=true

PR #186, opened 2022-10-03
Fix sign in lower level maximization QP

### LINKED ISSUE #182: Possible Bug with Maximization in Lower Level
Hi all, 
great package, it really makes life a lot easier!
I noticed some strange behavior when the lower level problem is a maximization. See attached a MWE illustrating this, it is based on: Siddiqui S, Gabriel SA (2013). An sos1-based approach for solving mpecs with a natural gas market applica-
tion. Networks and Spatial Economics 13(2):205–227.
Since, in the original publication there are two lower level players, I reformulated both lower level problems into a single one (with equivalent KKTs).
When minimizing the negative of the lower objective, results are as expected. However running the same code but switching to maximizing the actual objective gives wrong results. 

The MWE is: 

```julia
using JuMP, BilevelJuMP, Ipopt

F = [1,2]
c = Dict(1=>1, 2=>1)
C = 1
a = 13
b = 1

model = BilevelModel(Ipopt.Optimizer, mode = BilevelJuMP.ProductMode())

@variable(Lower(model), q[F] >= 0)
@variable(Upper(model), Q >= 0)

@objective(Upper(model), Max, ((a-b * (q[1] + q[2] + Q)) * Q - C*Q) )

@objective(Lower(model), Min, -((a-b * (q[1] + q[2] + Q)) * q[1] - C*q[1] + (a-b * (q[1] + q[2] + Q)) * q[2] - C*q[2] + b*q[1]*q[2]) )

optimize!(model)

@assert isapprox(value.(Q), 6; atol=1e-6)
@assert isapprox(value.(q).data, [2,2]; atol=1e-6) 
```

Edit: In an earlier version, there has been a copy/paste issue, this is now corrected

joaquimg:
Close #182
Replace #179 
- [ ] Needs tests

### DIFF
--- a/.github/workflows/documentation.yml
+++ b/.github/workflows/documentation.yml
@@ -8,6 +8,8 @@ on:
 jobs:
   build:
     runs-on: ubuntu-latest
+    env:
+      GKSwstype: nul
     steps:
       - uses: actions/checkout@v2
       - uses: julia-actions/setup-julia@latest
--- a/.github/workflows/documentation.yml
+++ b/.github/workflows/documentation.yml
@@ -15,8 +15,8 @@ jobs:
           # Build documentation on Julia 1.6
           version: '1.6'
       - name: Install dependencies
-        run: julia --project=docs/ -e 'using Pkg; Pkg.add(name="Dualization", version="0.5.5")'
-             #; Pkg.develop(PackageSpec(path=pwd())); Pkg.instantiate()'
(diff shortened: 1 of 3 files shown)


joaquimg:
@odow has any idea why docs keep failing?
Dualization 0.5.5 is released for hours.

joaquimg:
docs failed, but after some time worked again, probably a delay in General registry update

joaquimg:
replaced by #188

Issue #163, opened 2022-03-05
Update to newest Dualization.jl that considers constrained variables


joaquimg:

LukasBarner:
What needs to be done on this end?
I think this should greatly decrease solution times for problems with many variable bounds...
For quite a few examples, it already seems to work fine. 
Some of the tests still seem to fail, but this may also be for other reasons...

joaquimg:
I am not 100% sure this will change considerably the solution times.

Right now, BilevelJuMP is only able to write a KKT from a Dualization.jl Standard form.
To support this, we need to be able to handle Compact form.
The main missing piece is the second complementarity condition in: https://jump.dev/Dualization.jl/dev/manual/#KKT-3
but there is some bookkeeping to be done around.

LukasBarner:
So it would be required to add complements for the inequality stationarity conditions related to primal bounded variables, right? My guess that this would speed things up was because one dual variable per bound could actually be cut out then

LukasBarner:
Think I may have found a way for this without too much unnecessary bookkeeping: https://github.com/joaquimg/BilevelJuMP.jl/pull/189. 
Relevant are only the last three commits (the rest is the iterative stuff, just wanted to make sure it also works for that..). 
Do you think this might be worth further efforts?

LukasBarner:
PS: sorry for the spam here and on Dualization.jl the last few days :D

PR #189, opened 2022-10-05
Consider variable bounds

LukasBarner:
This is a first attempt to specifically consider variable bounds via complements. 
It should reduce solution times, since one dual variable per variable bound can be saved.
This PR also includes previous commits from https://github.com/joaquimg/BilevelJuMP.jl/pull/184 to check it also works with the iterative ProductMode.
@joaquimg could you take a look at the commits Oct. 5th and later?
Think this might be an easy way for https://github.com/joaquimg/BilevelJuMP.jl/issues/163. 
I believe it should work well except for aggregation and conic. 
Sorry for the spam the last few days :D

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -8,6 +8,7 @@ Dualization = "191a621a-6537-11e9-281d-650236a99e60"
 JuMP = "4076af6c-e467-56ae-b986-b466b2749572"
 LinearAlgebra = "37e2e46d-f89d-539d-b4ee-838fcccc9c8e"
 MathOptInterface = "b8f27783-ece8-5eb3-8dc8-9495eed66fee"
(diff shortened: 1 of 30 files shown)


LukasBarner:
True, this is more about if you think that https://github.com/joaquimg/BilevelJuMP.jl/pull/189/commits/b928983b9d76a224392e5b5cb102cd74e0b7426b is a reasonable approach in general. I don't think anything from this PR should really be merged...

LukasBarner:
sorry for the mess :D

joaquimg:
Ok, the general approach does seem reasonable, I´d gladly accept a PR.
Remember to:
 - add only the code that is required for the PR to accelerate the review process.
 - follow the style in general
 - add tests
 With those, the review process will be much faster

Issue #185, opened 2022-09-27
Variable bounds not included in list of constraint indices


LukasBarner:
Currently, model.ctr_lower does not include variable bound references: 
```julia
using JuMP, BilevelJuMP, Ipopt
model = BilevelModel(Ipopt.Optimizer, mode = BilevelJuMP.ProductMode())
@variable(Upper(model), x)
@variable(Lower(model), y>=0)
@objective(Upper(model), Min, -x + 4y)
@constraint(Upper(model), y + 2x <= 9)
@objective(Lower(model), Min, -x - y )
@constraint(Lower(model), x + y <= 8)
@constraint(Lower(model),  x <= 4)
optimize!(model)
model.ctr_lower
# Dict{Int64, ConstraintRef} with 2 entries:
#   4 => x ≤ 4.0
#   3 => x + y ≤ 8.0
model.lower_dual_to_sblm
# MathOptInterface.Utilities.IndexMap with 7 entries:
#   VariableIndex(3)                                                    => VariableIndex(3)
#   VariableIndex(1)                                                    => VariableIndex(4)
#   VariableIndex(2)                                                    => VariableIndex(5)
#   ConstraintIndex{VariableIndex, LessThan{Float64}}(2)                => ConstraintIndex{VariableIndex, LessThan{Float64}}(5)
#   ConstraintIndex{VariableIndex, LessThan{Float64}}(1)                => ConstraintIndex{VariableIndex, LessThan{Float64}}(4)
#   ConstraintIndex{VariableIndex, GreaterThan{Float64}}(3)             => ConstraintIndex{VariableIndex, GreaterThan{Float64}}(3)
#   ConstraintIndex{ScalarAffineFunction{Float64}, EqualTo{Float64}}(1) => ConstraintIndex{ScalarAffineFunction{Float64}, EqualTo{Float64}}(1)
model.lower_primal_dual_map.primal_con_dual_var
# Dict{MathOptInterface.ConstraintIndex, Vector{MathOptInterface.VariableIndex}} with 3 entries:
#   ConstraintIndex{ScalarAffineFunction{Float64}, LessThan{Float64}}(2) => [VariableIndex(2)]
#   ConstraintIndex{VariableIndex, GreaterThan{Float64}}(2)              => [VariableIndex(3)]
#   ConstraintIndex{ScalarAffineFunction{Float64}, LessThan{Float64}}(1) => [VariableIndex(1)]
```
[...]

PR #191, opened 2022-10-06
Consider variable bounds clean

LukasBarner:
Sorry for the mess on https://github.com/joaquimg/BilevelJuMP.jl/pull/189, I'm really not experienced in this at all :D 

This is a clean version including all old tests now for both consider_constrained_variables true and false. 
Still some failing tests for "Bilevel Conic JuMP NLP" and "Sum aggregation_group".
Since some tests in these testsets failed also on master (for me), "Bilevel Conic JuMP NLP" does not seem to be a problem. However I'm not sure about "Sum aggregation_group"...

I guess it could make sense to limit aggregation_group when working with consider_constrained_variables ? 

Please let me know if there is anything in terms of style etc. so I can improve.

### DIFF
--- a/src/jump.jl
+++ b/src/jump.jl
@@ -583,24 +583,9 @@ function JuMP.optimize!(
         copy_names = model.copy_names,
         pass_start = model.pass_start,
         consider_constrained_variables = consider_constrained_variables,
+        bilevel_model = model,
     )
 
-    # pass additional info (hints - not actual problem data)
-    # for lower level dual variables (start, upper hint, lower hint)
-    for (idx, info) in model.ctr_info
-        if haskey(model.ctr_lower, idx)
-            ctr = model.ctr_lower[idx]
-            # Only pass dual variable info if duals should exist. 
-            # This is not the case if constrained variables are considered during dualization: 
-            ctr_idx = JuMP.index(ctr)
-            F = MOI.get(lower, MOI.ConstraintFunction() , ctr_idx)
-            if !consider_constrained_variables || !(isa(F, MOI.VariableIndex) || isa(F, MOI.VectorOfVariables))
-                pre_duals = lower_primal_dual_map.primal_con_dual_var[ctr_idx] # vector
-                duals = map(x -> lower_dual_to_sblm[x], pre_duals)
-                pass_dual_info(single_blm, duals, info)
-            end
-        end
-    end
     # pass lower & upper level primal variables info (upper, lower)
     for (idx, info) in model.var_info
         if haskey(model.var_lower, idx)
@@ -612,6 +597,7 @@ function JuMP.optimize!(
         end
         pass_primal_info(single_blm, var, info)
     end
+    
     if length(bilevel_prob) > 0
         print_lp(single_blm, bilevel_prob, file_format)
     end
--- a/src/moi.jl
+++ b/src/moi.jl
@@ -433,6 +433,7 @@ function build_bilevel(
     copy_names::Bool = false,
     pass_start::Bool = false,
     consider_constrained_variables::Bool = false,
+    bilevel_model = BilevelModel(), #BilevelModel
 )
 
     # Start with an empty problem
@@ -539,6 +540,41 @@ function build_bilevel(
         pass_names(m, lower_dual, lower_dual_idxmap)
     end
 
+    # pass additional info (hints - not actual problem data)
+    # for lower level dual variables (start, upper hint, lower hint)
+    for (idx, info) in bilevel_model.ctr_info
+        if haskey(bilevel_model.ctr_lower, idx)
+            ctr = bilevel_model.ctr_lower[idx]
+            # Only pass dual variable info if duals should exist. 
+            # This is not the case if constrained variables are considered during dualization: 
+            ctr_idx = JuMP.index(ctr)
+            F = MOI.get(lower, MOI.ConstraintFunction() , ctr_idx)
+            if !consider_constrained_variables || !(isa(F, MOI.VariableIndex) || isa(F, MOI.VectorOfVariables))
+                pre_duals = lower_primal_dual_map.primal_con_dual_var[ctr_idx] # vector
+                duals = map(x -> lower_dual_idxmap[x], pre_duals)
+                pass_dual_info(m, duals, info)
+            end
+        end
+    end
+
+    if !consider_constrained_variables
+        # pass dual starts to single variable bounds, they are not part of bilevel_model.ctr_lower
+        for var in all_variables(bilevel_model.lower)
+            if has_lower_bound(var)
+                start_val = dual_start_value(LowerBoundRef(var))
+                lower_dual_bound_var = lower_primal_dual_map.primal_con_dual_var[JuMP.index(LowerBoundRef(var))][1] 
+                sblm_dual = lower_dual_idxmap[lower_dual_bound_var]
+                MOI.set(m, MOI.VariablePrimalStart(), sblm_dual, start_val)
+            end
+            if has_upper_bound(var)
+                start_val = dual_start_value(UpperBoundRef(var))
+                lower_dual_bound_var = lower_primal_dual_map.primal_con_dual_var[JuMP.index(UpperBoundRef(var))][1] 
+                sblm_dual = lower_dual_idxmap[lower_dual_bound_var]
+                MOI.set(m, MOI.VariablePrimalStart(), sblm_dual, start_val)
+            end
+        end
+    end
+    
     #=
         Additional Optimiality conditions (to complete the KKT)
     =#
@@ -887,6 +923,12 @@ function add_complement(
     f_dest = MOIU.map_indices.(Ref(idxmap_primal), f)
     new_f = MOIU.operate(-, T, f_dest, slack)
     equality = MOIU.normalize_and_add_constraint(m, new_f, MOI.EqualTo(zero(T)))
+
+    slack_start = MOIU.eval_variables(
+        x -> nothing_to_nan(MOI.get(m, MOI.VariablePrimalStart(), x)),
+        f_dest,
+    )
+    MOI.set(m, MOI.VariablePrimalStart(), slack, slack_start)
 
     dual = idxmap_dual[v]
     c1 = MOI.add_constraint(
@@ -1249,6 +1291,8 @@ function add_complement(
     is_tight = false
     has_start = false
 
+    dual = idxmap_dual[v]
+
     if mode.with_slack
         slack, slack_in_set = MOI.add_constrained_variable(m, s)
     end
@@ -1265,7 +1309,7 @@ function add_complement(
             f_dest,
         )
         if !isnan(val)
-            is_tight = abs(val) < 1e-8
+            is_tight = abs(val) < abs(nothing_to_nan(MOI.get(m, MOI.VariablePrimalStart(), dual)))
             has_start = true
         end
     end
@@ -1279,15 +1323,13 @@ function add_complement(
         end
     end
 
-    dual = idxmap_dual[v]
     v_bounds = get_bounds(dual, mode.cache.map, mode.dual_big_M)
 
     bin = MOI.add_variable(m)
     if pass_start && has_start && is_tight
-        MOI.set(m, MOI.VariablePrimalStart(), bin, 1.0)
-        MOI.set(m, MOI.VariablePrimalStart(), dual, 0.0)
-    else
         MOI.set(m, MOI.VariablePrimalStart(), bin, 0.0)
+    else
+        MOI.set(m, MOI.VariablePrimalStart(), bin, 1.0)
     end
 
     s1 = flip_set(s)
@@ -1319,7 +1361,7 @@ function add_complement(
         -Mv,
     )
 
-    c1 = MOIU.normalize_and_add_constraint(m, f1, s2)
+    c1 = MOIU.normalize_and_add_constraint(m, f1, s1)
     c2 = MOIU.normalize_and_add_constraint(m, f2, s2)
     c3 = MOI.add_constraint(m, bin, MOI.ZeroOne())
 


src/moi.jl
```diff
@@ -255,6 +255,52 @@ end
...
+    dual_set = MOI.get(dual_model, MOI.ConstraintSet(), dual_con)
+    if MOI.constant(dual_set) > 0
+        dual_func.constant = dual_func.constant - MOI.constant(dual_set)
+    elseif MOI.constant(dual_set) < 0
+        dual_func.constant = dual_func.constant + MOI.constant(dual_set)
+    end
+    return Complement(false, primal_con, dual_func, set_with_zero(dual_set), primal_variable)
+end
+
+function get_variable_complement(primal_model, dual_model, primal_con::MOI.ConstraintIndex{Fp,Sp}, dual_con::MOI.ConstraintIndex{Fd,Sd}) where {Fp<:MOI.VariableIndex,Sp<:MOI.LessThan{T},Fd,Sd<:MOI.LessThan{T}} where T
```
joaquimg, line 278: these two functions look the same

src/moi.jl
```diff
@@ -428,8 +475,32 @@ function build_bilevel(
...
+            end
+        end
+
     else # strong duality
-        add_strong_duality(mode, m, lower_primal_obj, lower_dual_obj, lower_idxmap, lower_dual_idxmap)
+        if !consider_constrained_variables || contains_only_scalar_sets(lower)
+            add_strong_duality(mode, m, lower_primal_obj, lower_dual_obj, lower_idxmap, lower_dual_idxmap)
+        else
+            error("Only scalar sets in lower level allowed with consider_constrained_variables.")
+        end
```
joaquimg, line 503: I dont think any of this is needed

test/jump.jl
```diff
@@ -1,6 +1,6 @@
-jump_01vec(optimizer, mode = BilevelJuMP.SOS1Mode(), config = Config()) = _jump_01(optimizer, true, mode, config)
-jump_01(optimizer, mode = BilevelJuMP.SOS1Mode(), config = Config()) = _jump_01(optimizer, false, mode, config)
-function _jump_01(optimizer, vectorized::Bool, mode, config)
+jump_01vec(optimizer, mode = BilevelJuMP.SOS1Mode(), consider_constrained_variables = consider_constrained_variables, config = Config()) = _jump_01(optimizer, true, mode, consider_constrained_variables, config)
```
joaquimg, line 1: this should be a field in config.
Also, I am not sure this should be run in all tests.

test/runtests.jl
```diff
@@ -120,283 +120,313 @@ end
...
-        jump_06_sv(solver.opt, solver.mode)
-        # jump_07(solver.opt, solver.mode, CONFIG_2)
-        jump_08(solver.opt, solver.mode, CONFIG_3_start)
-        jump_09a(solver.opt, solver.mode)
-        # jump_09b(solver.opt, solver.mode)
-        jump_11a(solver.opt, solver.mode)
-        jump_11b(solver.opt, solver.mode)
-        # jump_12(solver.opt, solver.mode)
-        # jump_14(solver.opt, solver.mode)
+    for consider_constrained_variables in [true, false]
```
joaquimg, line 123: I dont think this should be run in all tests that would lead to run everything twice.
I think we should select a subset.

src/moi.jl
```diff
@@ -255,6 +255,52 @@ end
 accept_vector_set(::ProductMode{T}, ::Complement) where T = nothing
 accept_vector_set(::MixedMode{T}, ::Complement) where T = nothing
 
+function get_variable_complement(primal_model, dual_model, primal_con, dual_con) 
+    error("An internal error with variable complements occurred. Likely, your problem type does not yet support consideration of constrained variables.")
+end
+
+function get_variable_complement(primal_model, dual_model, primal_con::MOI.ConstraintIndex{Fp,Sp}, dual_con::MOI.ConstraintIndex{Fd,Sd}) where {Fp<:MOI.VariableIndex,Sp<:MOI.GreaterThan{T},Fd,Sd<:MOI.GreaterThan{T}} where T
```
joaquimg, line 262: any reason for not supporting other sets like vector sets?
you can follow the code in get_canonical_complement, as you have been doing

joaquimg (changes requested):
Good start!
- We need this to work with vector sets.
- Aggregation groups should just work (we can leave it for a little later)
- I don't think we should test everything twice in the final version (we can leave it there while we finish the code.
- Bilevel Conic requires vector sets

joaquimg:
Note that CI will only run if you fix conflicts with master.

src/moi.jl
LukasBarner: Yes, this was an old relict as I wasn't sure about constants etc in the beginning. Now, one is deleted and the other is defined for union of lesser/greater. Equalities still throw an error.

src/moi.jl
LukasBarner: true, this threw occasional errors earlier. Found out that this was related to missing dual variables for which start values could not be passed. 
This is now resolved, since hints etc. are no longer passed if the dual to bounds does not exist in the first place...

test/jump.jl
LukasBarner: Left this for now, as I'm not sure which tests to select best etc. 
Personally, I don't think that running tests twice adds a huge amount of time, as almost everything should be compiled already and the problems are fairly small. 
On my machine, running tests like this adds about 6 seconds (less than 7%)

LukasBarner: But I agree a field in config would be better, will change this...

test/runtests.jl
LukasBarner: See above

test/jump.jl
LukasBarner: I could also do the subset, but I'm not sure which ones to select...
Do you have a preference?

src/moi.jl
LukasBarner: left this out because I did not really feel certain. 
Now added, can you check if it is correct? 
Also not sure about the todo (see code for the new function)

LukasBarner:
can you tell me if I resolved the conflicts correctly?

src/jump.jl
```diff
@@ -681,6 +687,38 @@ function pass_primal_info(single_blm, primal, info::BilevelVariableInfo)
...
+        return
+    else
+        pre_duals = lower_primal_dual_map.primal_con_dual_var[ctr_idx] # vector
+        duals = map(x -> lower_dual_to_sblm[x], pre_duals)
+        pass_dual_info(single_blm, duals, info)
+    end
+    return
+end
+
+function pass_necessary_dual_info(
```
joaquimg, line 708: same function twice?

src/jump.jl
```diff
@@ -590,10 +592,14 @@ function JuMP.optimize!(
             ctr = model.ctr_lower[idx]
             # this fails for vector-constrained variables due dualization 0.3.5
             # because of constrained variables that change the dual
-            pre_duals =
-                lower_primal_dual_map.primal_con_dual_var[JuMP.index(ctr)] # vector
-            duals = map(x -> lower_dual_to_sblm[x], pre_duals)
-            pass_dual_info(single_blm, duals, info)
+            pass_necessary_dual_info(
```
joaquimg, line 595: why a function if a simple conditional would suffice?

src/moi.jl
```diff
@@ -289,6 +289,88 @@ end
...
+end
+
+function get_variable_complement(
+    primal_model,
+    dual_model,
+    primal_con::MOI.ConstraintIndex{Fp,Sp},
+    dual_con::MOI.ConstraintIndex{Fd,Sd},
+) where {
+    Fp<:MOI.VariableIndex,
+    Sp<:Union{MOI.LessThan{T},MOI.GreaterThan{T}},
```
joaquimg, line 305: SCALAR_SETS ?

src/moi.jl
```diff
@@ -289,6 +289,88 @@ end
...
+    Sd<:Union{MOI.LessThan{T},MOI.GreaterThan{T}},
+} where {T}
+    primal_variable =
+        MOI.get(primal_model, MOI.ConstraintFunction(), primal_con)
+    primal_set = MOI.get(primal_model, MOI.ConstraintSet(), primal_con)
+
+    @assert MOI.constant(primal_set) == 0 "Unexpected variable bound"
+
+    dual_func = MOI.get(dual_model, MOI.ConstraintFunction(), dual_con)
+    dual_set = MOI.get(dual_model, MOI.ConstraintSet(), dual_con)
```
joaquimg, line 316: you need MOI.copy just like in get_canonical_complement

src/moi.jl
```diff
@@ -289,6 +289,88 @@ end
...
+        MOI.get(primal_model, MOI.ConstraintFunction(), primal_con),
+    )::Fp
+    primal_set =
+        MOI.copy(MOI.get(primal_model, MOI.ConstraintSet(), primal_con))::Sp
+
+    dual_func =
+        MOI.copy(MOI.get(dual_model, MOI.ConstraintFunction(), dual_con))
+    dual_set = MOI.copy(MOI.get(dual_model, MOI.ConstraintSet(), dual_con))
+
+    # Do we have a todo here as in function get_canonical_complement(primal_model, map, ci::CI{F,S}) where {F, S<:VECTOR_SETS} ??
```
joaquimg, line 347: do what?

joaquimg (changes requested):

test/jump.jl
joaquimg: if the overhead is only 7% for running twice, this would indeed be fine

src/jump.jl
LukasBarner: sorry, thought I already replied to these... 
now a conditional

src/jump.jl
LukasBarner: done, see above

src/moi.jl
LukasBarner: Would prefer this to throw an error in case of equalities (i.e. if something unexpected happens). 
Could also go with SCALAR_SETS for code readability in case you prefer and do @assert inside the function?

src/moi.jl
LukasBarner: There was the following todo in the code of get_canonical_complement (note that I commented one additional line out):
```julia
    # dim = MOI.dimension(set)
    # vector sets have no constant
    # for i in 1:dim
    #     func.constant[i] = Dualization.set_dot(i, set, T) *
    #         Dualization.get_scalar_term(primal_model, ci, i)
    # end
    # todo - set dot on function
```

src/moi.jl
LukasBarner: Now done, but only for the dual_func. This is the only necessary one, right? 
I'm asking because the copy is done for sets etc. in other functions (but that should be handled by set_with_zero, right? 

PS: This was a big one, so I'm a bit surprised it didn't affect testing...

LukasBarner:
Quick heads up: I just noticed that there are still some problems with setting correct start values. However, considering https://github.com/joaquimg/BilevelJuMP.jl/issues/185, this PR is still an overall improvement. 
In comparison to before, ProductMode works for constrained variables now (probably also a few others). 
I can still confirm issues with at least SOS1Mode and FortunyAmatMcCarlMode. There are a few things that need additional adjustment for the variable complements here. 
However, there are some more things that do not work as expected even when not using the newly added variable complements...
I have a larger sample problem for which I can compute a feasible starting point (by finding a solution to the lower level with fixed upper level variables). When passing it in ProductMode, I get a feasible starting point to the single level model. However, when I use FortunyAmatMcCarlMode, starting points are not accepted (not an issue of different solver tolerances etc., violations are orders of magnitude larger). 
We could either take care of investigating the start value thing (for both newly added and existing functionality) in this PR, or move either both or one to a new issue/PR. I think I would prefer a new issue/PR for both, what do you think @joaquimg ?

joaquimg:
We have to figure out what the issues are with SOS1Mode and FortunyAmatMcCarlMode.
They may be more fundamental things that are working in ProductMode just by chance.

Start values should be passed, this should not be hard.

Start values are frequently ignored by MIP solvers (used in FortunyAmatMcCarlMode). How do you know they are not accepted? MIP solvers usually require starting points to absolutely all variables involved.

If things don't work for existing functionality, we should have issues to discuss them separately.

This PR should not be blocked by things that were not previously working, but it should be blocked by things that previously worked.

LukasBarner:
Quick disclaimer: As I said, I'm pretty new to this and not 100% sure about everything. In case it is nonsense, please excuse the effort of reading through all of it...

First a few replies to the text above: 
> Start values should be passed, this should not be hard.

100% agree with the first part (especially since it will be a heavily used functionality), but not so sure that this is a trivial issue (more in the second part...)

> Start values are frequently ignored by MIP solvers (used in FortunyAmatMcCarlMode). How do you know they are not accepted? 

For smaller examples, start values are sometimes accepted. Using Gurobi, I get notifications when start values violate constraints (i.e. constraint X is violated by 60.08). This is usually the case for larger models. 

> MIP solvers usually require starting points to absolutely all variables involved.

My code in general does set all start values correctly, this can be verified when switching to ProductMode with Ipopt, where initial point primal infeasibility of the single level model constraints is about e-7. Note that this only holds if `consider_constrained_variables = true`, otherwise the duals to variable bounds are not being set correctly (more on this also below). Also, I do not believe that ProductMode is working "by chance", things there should be largely correct...

Now a few remarks to the existing functionality and why I'm not sure starting values are a trivial issue. 

Currently, dual start values to lower level variable bounds are not set. Some exceptions exist, such as FortunyAmatMcCarlMode, where duals are sometimes set to zero (which tbh I'm not sure is actually correct: If the constraint was tight, why do we set the dual variable start to 0? ):
https://github.com/joaquimg/BilevelJuMP.jl/blob/565c0ef6d5fd07ae7ff558bbf2466b87e815caf9/src/moi.jl#L1181-L1185

As mentioned in https://github.com/joaquimg/BilevelJuMP.jl/issues/185, model.ctr_lower does not contain variable bounds currently (see the issue for an MWE demonstrating this). I believe, that in the following lines, this leads to their dual starts not being passed at all (exception above): 
https://github.com/joaquimg/BilevelJuMP.jl/blob/2d25399b307031dfd5de46153fa092b5469e3ffe/src/jump.jl#L515-L524

Furhter, I believe another reason why I have failing tests on larger models with start values in FortunyAmatMcCarlMode could be: 
https://github.com/joaquimg/BilevelJuMP.jl/blob/565c0ef6d5fd07ae7ff558bbf2466b87e815caf9/src/moi.jl#L1158-L1167
Potentially, numerical imprecision and solver tolerances may lead to deviations larger than 1e-8, falsely classifying a constraint as not tight. I know this to be true in my case, but I'm not sure its the only issue. 
Another way to set the start values of binaries could be to take a look at the dual start values and check which deviation from zero of either the function or the dual variable is bigger. 
A problem though is that the dual starts are not yet available in this part of the code...

These lines are actually also a problem for the newly introduced variable complements: 
https://github.com/joaquimg/BilevelJuMP.jl/blob/565c0ef6d5fd07ae7ff558bbf2466b87e815caf9/src/moi.jl#L1158-L1167
Since duals do not yet have start values set, `val` may evaluate to NaN for the new variable complements despite all starts being set correctly... 

For these reasons, I think it might make sense to already pass start values of duals in the 'build_bilevel' function before starting to set up complementary slackness (and to make sure that all starts are actually passed of course :D). 

As you can see, there are a few issues that can be discussed, but I'm afraid it would certainly blow up this PR...
This is why I would prefer to make start values a new issue once this PR is ready. 

PS: I have not taken a deeper dive into SOS1, but I figure similar issues are going on there...

LukasBarner:
> We have to figure out what the issues are with SOS1Mode and FortunyAmatMcCarlMode.
> They may be more fundamental things that are working in ProductMode just by chance.

You were right here! There is something more fundamental... 
I already caught one problem when handling scalar set constants, let me check if there is another. 
Will correct soon...

LukasBarner:
Okay, think I have figured this one out...
Also have everything that is written below on a new branch (which I could either make a new start value PR, or include to this one, whichever you prefer).
The codes mentioned below are also tested on a larger model and work as expected there.
I have split things up into stuff related to this PR and things that are also valid for the current master. 
Sry for the long text... 

## Things that are fundamentally wrong (but easy to fix)
### Obvious mistake in this PR: 
First the general mistake, these lines I have written are obvious nonsense:
https://github.com/LukasBarner/BilevelJuMP.jl/blob/422b4060f28006abdfb69d3921475670b28d8743/src/moi.jl#L317-L321
```julia
if MOI.constant(dual_set) > 0
    dual_func.constant = dual_func.constant - MOI.constant(dual_set)  
elseif MOI.constant(dual_set) < 0
    dual_func.constant = dual_func.constant + MOI.constant(dual_set)
end
```
They should instead be: 
```julia
dual_func.constant = dual_func.constant - MOI.constant(dual_set)
```
### Passing of dual starts: 
Currently, dual starts are passed after the bilevel problem is built: 
https://github.com/joaquimg/BilevelJuMP.jl/blob/565c0ef6d5fd07ae7ff558bbf2466b87e815caf9/src/jump.jl#L571-L598
This is fine if we do not have the new "variable complements". However, the new "variable complements" also have dual variables inside the ConstraintFunction, and some start value related functionalities (i.e. setting starting values of binary variables) rely upon evaluating starts:  https://github.com/joaquimg/BilevelJuMP.jl/blob/565c0ef6d5fd07ae7ff558bbf2466b87e815caf9/src/moi.jl#L1158-L1167
Since the dual start values are not yet passed for the variable complements, this tanks the whole warmstart procedure for MIP solvers that require fully feasibly start values (IMO usually the case).
To make warmstarts work here, we have to move passing dual starts into the build_bilevel(...) function. Luckily, this is just copy-pasting and works fine. 
We need to pass the bilevel model as an argument to build_bilevel and then copy this piece https://github.com/joaquimg/BilevelJuMP.jl/blob/565c0ef6d5fd07ae7ff558bbf2466b87e815caf9/src/jump.jl#L586-L609
and paste it here: https://github.com/joaquimg/BilevelJuMP.jl/blob/565c0ef6d5fd07ae7ff558bbf2466b87e815caf9/src/moi.jl#L460-L464 prior to the complement part (Additional Optimiality conditions). 

## Start values for SOS1 and FortunyAmatMcCarl
Now to the issue with start values for SOS1 and FortunyAmatMcCarl: 
### SOS1 Starts:
Currently, no start values are set for the newly added slack variables here: 
https://github.com/joaquimg/BilevelJuMP.jl/blob/565c0ef6d5fd07ae7ff558bbf2466b87e815caf9/src/moi.jl#L765-L803
This also holds if we do not consider constrained variables during dualization. However, it can easily be fixed by something like this inside the add_complement(...) function: 
```julia
slack_start = MOIU.eval_variables(
    x -> nothing_to_nan(MOI.get(m, MOI.VariablePrimalStart(), x)),
    f_dest,
)
MOI.set(m, MOI.VariablePrimalStart(), slack, slack_start)
```
Now, SOS1 start values work as expected!

### FortunyAmatMcCarl Starts:
Here, binary starts (and duals to zero) were mixed up: 
https://github.com/joaquimg/BilevelJuMP.jl/blob/565c0ef6d5fd07ae7ff558bbf2466b87e815caf9/src/moi.jl#L1182-L1187
Which I believe should instead be: 
```julia
if pass_start && has_start && is_tight
    MOI.set(m, MOI.VariablePrimalStart(), bin, 0.0)
else
    MOI.set(m, MOI.VariablePrimalStart(), bin, 1.0)
    MOI.set(m, MOI.VariablePrimalStart(), dual, 0.0)
end
```
Since MIP solvers usually require a complete set of starts, I believe that user provided duals should not be over-written, so 
```julia
if pass_start && has_start && is_tight
    MOI.set(m, MOI.VariablePrimalStart(), bin, 0.0)
else
    MOI.set(m, MOI.VariablePrimalStart(), bin, 1.0)
end
```
might even be a better solution... 

Finally, this line can make problems when providing warmstarts:
https://github.com/joaquimg/BilevelJuMP.jl/blob/565c0ef6d5fd07ae7ff558bbf2466b87e815caf9/src/moi.jl#L1164
a) For larger models, it is likely that some numerical imprecision occurs. 
b) If warmstarts are obtained by solving the lower level for fixed upper level variables, it is not guaranteed that solver tolerances are tighter than 1e-8 (Gurobi for example uses 1-6 by default). 

One workaround would be to make the tolerance an argument that can be specified by the user. 
But since MIP solvers usually require a complete set of warmstarts, there is an easier solution. We can just compare `val` to the dual variable (this should be more numerically stable and eliminates user defined tolerances). 
Something like this does work (only if dual starts are already passed in the build_bilevel(...) function though): 
```julia
is_tight = abs(val) < abs(nothing_to_nan(MOI.get(m, MOI.VariablePrimalStart(), dual)))
```

joaquimg:
Your comments seem reasonable indeed.
You can either update this PR or propose new ones, I am very keen to see them.
Remember the key point:
Small PR's are much easier to review and, consequently, quicker to integrate. We don't want to add bugs by accident.
Large PR's are much harder and slower to review since I have to go through each line and think about possible issues.

LukasBarner:
Ok, decided to keep them in this PR. 
I understand that this makes reviews more difficult, but there are synergies between the consider_constrained_variables case and the start values.
This should now also fix https://github.com/joaquimg/BilevelJuMP.jl/issues/185. 
I think warm starts (especially for the integer methods) are extremely useful and will be frequently used, so fixing them should have a high priority. In case this review will take significantly longer to get ready (or fail due to other problems), I could also do an "interim PR" that ignores some of the consider_constrained_variables functionality to make start values available quicker. What do you think?

PR #195, opened 2023-02-27
Test with Alpine

joaquimg:
Running alpine with Ipopt+HiGHS.

Most cases fail.

1 - Many seem sem to be trying to add a quadratic constraint to HiGHS:
```julia
 Constraints of type MathOptInterface.ScalarQuadraticFunction{Float64}-in-MathOptInterface.GreaterThan{Float64} are not supported by the solver.

  If you expected the solver to support your problem, you may have an error in your formulation. Otherwise, consider using a different solver.

  The list of available solvers, along with the problem types they support, is available at https://jump.dev/JuMP.jl/stable/installation/#Supported-solvers.
  Stacktrace:
    [1] error(s::String)
      @ Base .\error.jl:35
    [2] _moi_add_constraint(model::MathOptInterface.Utilities.CachingOptimizer{MathOptInterface.Bridges.LazyBridgeOptimizer{HiGHS.Optimizer}, MathOptInterface.Utilities.UniversalFallback{MathOptInterface.Utilities.Model{Float64}}}, f::MathOptInterface.ScalarQuadraticFunction{Float64}, s::MathOptInterface.GreaterThan{Float64})
```

I would expect that quadratic constraints are forwarded to Ipopt as non-linear constraints.

2 - Another common error is:
```julia
  Got exception outside of a @test
  Result index of attribute MathOptInterface.ObjectiveValue(1) out of bounds. There are currently 0 solution(s) in the model.
  Stacktrace:
    [1] check_result_index_bounds
      @ C:\Users\joaquimgarcia\.julia\packages\MathOptInterface\NCblk\src\attributes.jl:207 [inlined]
    [2] get
      @ C:\Users\joaquimgarcia\.julia\packages\HiGHS\QRk9O\src\MOI_wrapper.jl:1880 [inlined]
    [3] get(b::MathOptInterface.Bridges.LazyBridgeOptimizer{HiGHS.Optimizer}, attr::MathOptInterface.Bridges.ObjectiveFunctionValue{MathOptInterface.ScalarAffineFunction{Float64}})
```
Seems that Alpine is not handling some HiGHS fails well. Possibly due unbounded variables?

cc @harshangrjn @odow

### DIFF
--- a/test/Project.toml
+++ b/test/Project.toml
@@ -8,7 +8,12 @@ QuadraticToBinary = "014a38d5-7acb-4e20-b6c0-4fe5c2344fd1"
 Test = "8dfed614-e22c-5e08-85e1-65c5234f0b40"
 Pkg = "44cfe95a-1eb2-52ea-b672-e2afdf69b78f"
 
+Alpine = "07493b3f-dabb-5b16-a503-4139292d7dd4"
+HiGHS = "87dc4568-4c63-4d18-b0c0-bb2238e4078b"
+
 [compat]
 Ipopt = "=1.0.2"
 MibS_jll = "=1.1.3"
 QuadraticToBinary = "=0.4.0"
+
+Alpine = "=0.5.3"
--- a/test/jump.jl
+++ b/test/jump.jl
@@ -2888,14 +2888,14 @@ By: LukasBarner
 Siddiqui S, Gabriel SA (2013). An sos1-based approach for solving mpecs with a natural gas market applica-
 tion. Networks and Spatial Economics 13(2):205–227.
 """
-function jump_qp_lower_min(config = Config())
+function jump_qp_lower_min(optimizer, config = Config())
     F = [1,2]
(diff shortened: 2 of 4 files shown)


harshangrjn:
If the problem has convex quadratic constraints, then using Pavito + Highs  should work. Ipopt is used only for feasibility. Unbounded variables shouldn't be an issue as Alpine automatically applies large bounds on them and then tries to tighten their bounds. If you still have issues, share a working example where you can reproduce the issue.

harshangrjn:
@joaquimg Did using Pavito work like here: 
https://github.com/lanl-ansi/Alpine.jl/blob/8a3f9c56968dfbe49c4aa321e3bda47f2fb1c343/test/runtests.jl#L37
and 
https://github.com/lanl-ansi/Alpine.jl/blob/8a3f9c56968dfbe49c4aa321e3bda47f2fb1c343/test/test_algorithm.jl#L104

PR #238, opened 2026-03-22
Updating dual variable documentation page

biegelk:
This PR updates the documentation on the ["Dual variables of the lower level"](https://joaquimg.github.io/BilevelJuMP.jl/stable/tutorials/lower_duals/) page:

* Puts the basic definition and use of `DualOf` at the top of the page for convenience
* Reorganizes the example problem for more clarity
* Fixes mismatched variable names (mixed use of g1/gR1 and g2/gR2 --> all g1 and g2)
* Some copy-editing throughout

I wasn't able to test all the formatting, but I would be happy to make any fixes in case of mis-rendering.


joaquimg:
Thanks for the cleanup — the reorganization reads much better.

The docs build was failing, so I pushed a small fix to this branch (63712ab). The `make.jl` script `include`s every tutorial `.jl` file to test it, so the `DualOf` snippet that now sits at the top of the page was being executed before `using BilevelJuMP`:

```
LoadError: UndefVarError: `@constraint` not defined
  in expression starting at docs/src/tutorials/lower_duals.jl:9
```

(it also refers to `model`, `x` and `y`, which don't exist at that point).

I wrapped it in a commented ` ```julia ` block, which is the convention already used for non-executed snippets in `getting_started.jl` and `conic_lower.jl`. The rendered page is unchanged.

All checks pass now.

PR #244, opened 2026-08-20
Fix use of SCIP and Xpress in tests (rebased #223)

joaquimg:
Rebase of #223 (odow's `od/xpress`) onto current master, to get fresh CI signal.

#223 is currently `CONFLICTING` with master and its CI runs are from May 2025 with expired logs (HTTP 410), so its red checks were unreadable. This rebase resolves the conflict and re-runs CI.

Original commits by @odow, authorship preserved.

## Conflict resolution
Only conflict was the `ci.yml` matrix; resolved in favour of #223's version (`os × version` incl. `macos-latest`). Kept master's newer action versions (`checkout@v6`, `cache@v3`).

## What this changes
- Xpress test switched to `Xpress_jll` + `XPAUTH_XPR`, but **left commented out** (`# TODO(odow): FIXME`) — this is what turns the Windows job green: it stops attempting the `Pkg.build("Xpress")` that fails with `Unable to locate Xpress installation`.
- SCIP now runs on all platforms (was `Sys.islinux()`-gated).
- `Ipopt` compat loosened `=1.0.2` → `"1"`.

## Testing notes
- Full suite on Windows: 1461 passed, 1 errored.
- The 1 error is `libMibs-0.dll` failing to load — a **local Windows env issue**, not caused by this branch: in a clean env `MibS_jll` alone fails, while `MibS_jll` + SCIP succeeds (SCIP supplies a runtime DLL MibS needs). MibS is green on Linux CI.
- macOS dependency resolution verified for both `x86_64-macos` and `aarch64-macos`.

## Caveats for review
- Xpress coverage remains disabled — this makes that explicit rather than a red check.
- The `Ipopt` compat loosening affects the currently-green Linux jobs.
- Adds `macos-latest` × 2 versions as new job configs; note `macos-latest` is ARM64 while the matrix pins `arch: x64`.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


joaquimg:
## Why this is failing

The 6 test-job failures are **not** the Xpress change — they are the `Ipopt` compat loosening.

All 6 jobs fail identically with 1461 passed / 1 errored, in the MibS testset:

```
LoadError: InitError: could not load library ".../libMibs.so"
libBlis.so.0: cannot open shared object file: No such file or directory
```

`Ipopt = "1"` (was `=1.0.2`) lets `Ipopt_jll` float to `300.1400.1900`, which drags `Cbc_jll` / `Cgl_jll` / `Clp_jll` / `CoinUtils_jll` to builds ABI-incompatible with the pinned `MibS_jll` v1.1.3. On the passing `master`-based jobs, `Ipopt_jll` resolves to `300.1400.400` with a self-consistent COIN-OR stack.

I tried simply restoring `Ipopt = "=1.0.2"` here. **It is unsatisfiable** — the same PR makes SCIP run on all platforms, and `SCIP_jll` requires `Ipopt_jll >= 300.1400.1400`:

```
Ipopt_jll [9cc047cb] log:
├─restricted by compatibility requirements with Ipopt [b6b21f68] to versions: 3.13.2 - 300.1400.400
└─restricted by compatibility requirements with SCIP_jll [e5ac4fe4] to versions: 300.1400.1400 - 300.1400.1902 — no versions left
```

So the SCIP-everywhere change and the `MibS_jll` v1.1.3 pin are in direct conflict. Resolving it needs a newer `MibS_jll` built against the current COIN-OR stack, or keeping SCIP gated to Linux, or relaxing the `MibS_jll` pin. That is a design decision, so I have not guessed at it here.

Details in #245.

## Also

- `Xpress = "0.17"` should be `"0.18"` — `XPRESS_JL_LIBRARY` / `Xpress_jll` support landed in Xpress.jl v0.18, so the staged `xpress.jl` code could not work under the 0.17 pin.
- The Xpress portion itself looks correct and matches the [Xpress.jl](https://github.com/jump-dev/Xpress.jl/blob/master/.github/workflows/ci.yml) / [MOI solver-tests](https://github.com/jump-dev/MathOptInterface.jl/actions/workflows/solver-tests.yml) pattern — MOI runs Xpress on a plain `ubuntu-latest` with no license step, since Xpress.jl v0.18+ loads from `Xpress_jll`. It is just still gated off behind `# TODO(odow): FIXME`.

For #243 I have taken the narrow path instead: commented out the Windows Xpress block with a pointer to #245, leaving this PR's broader rework to be resolved separately.

joaquimg:
## Now green ✅

All checks pass. Four fixes, in order of what each one uncovered:

**1. `Ipopt` compat + SCIP gating.** Restored `Ipopt = "=1.0.2"` and moved SCIP back to a Linux-only runtime `Pkg.add`, dropping it as a hard test dep.

This was the actual cause of the `libBlis.so.0` MibS failures. `Ipopt = "1"` floated `Ipopt_jll` to `300.1400.1900`, dragging the COIN-OR stack to builds ABI-incompatible with the pinned `MibS_jll` v1.1.3. But simply restoring the pin is unsatisfiable while SCIP 0.12+ is a hard dep, since `SCIP_jll` needs `Ipopt_jll >= 300.1400.1400`.

I checked whether an older SCIP threads the needle: **SCIP 0.11.x does resolve** with the `Ipopt` pin (0.11.12 and 0.11.14 both OK; 0.12.0 / 0.12.5 / 0.12.8 all fail). But SCIP 0.11.14 then **segfaults on Windows**:

```
EXCEPTION_ACCESS_VIOLATION at SCIPprobFree ... libscip.dll
SCIPfreeTransform → SCIPfreeProb → SCIPfree → free_scip (finalizer)
```

So no single SCIP version works everywhere. Keeping it Linux-gated and installed at runtime keeps it out of the resolved environment on Windows/macOS entirely — which is what `master` already did, and it resolves to the same known-good stack as #243 (`Ipopt_jll 300.1400.400`, `Cbc_jll 200.1000.501`).

That makes a separate SCIP CI job unnecessary, and avoids splitting coverage reporting.

**2. `Xpress` compat 0.17 → 0.18**, since `XPRESS_JL_LIBRARY` / `Xpress_jll` support landed in Xpress.jl v0.18.

**3. Xpress licence wiring.** Added a CI step that writes `XPAUTH_XPR` to disk and exports `XPAUTH_PATH`, mirroring [MOI's `solver-tests.yml`](https://github.com/jump-dev/MathOptInterface.jl/blob/master/.github/workflows/solver-tests.yml). `xpress.jl` honours a pre-set `XPAUTH_PATH` and falls back to writing the licence itself, and the `include` is re-enabled behind a licence check instead of `# TODO(odow): FIXME`.

**The mechanism now works end to end** — the licence is written, found, and read by the solver. It surfaced this:

```
XpressError(8): Corrupt use. Xpress licensing error 10:
Your license expired on 28-feb-2026.
```

So the `XPAUTH_XPR` secret is ~6 months expired. That is a hard failure at `using Xpress`, which took down all four jobs, so an unusable licence is now treated the same as an absent one, with a warning. **Once the secret is renewed, Xpress starts being exercised again with no code change.**

On the community-licence idea: the PyPI `xpress` wheel ships only `LICENSE.txt` (legal text) and `_xpress.*.so` — no `xpauth.xpr` and no `libxprs`. `Xpress_jll` 9.4.3 ships no licence file either. FICO's Community Licence needs registering at [their site](https://www.fico.com/en/fico-xpress-community-license), so it can't be vendored from a package — a renewed secret is the path.

**4. Dropped macOS from the matrix.** I first fixed the arch (`macos-latest` is ARM64, so `x64` was wrong), which got macOS running — and it then failed 864 passed / 7 errored:

```
UndefVarError: `mibs` not defined in `MibS_jll`
```

`MibS_jll` installs on macOS aarch64 but ships no `mibs` executable. This branch's `test_mibs.jl` calls `MibS_jll.mibs` unconditionally; #243 adds the `MIBS_AVAILABLE` / `is_available()` guard that skips binary-dependent tests on such platforms. So macOS is left out for now with a comment — worth revisiting once #243's guard is on master.

Net result: same three configurations as before, all green, with the Xpress path live and waiting only on a valid licence.

joaquimg:
## The community licence works — you were right

`https://pypi.org/project/xpress` does bundle it, at `xpress/license/community-xpauth.xpr` (599 bytes, `hostid="any" platform="any"`). I had missed it earlier by grepping only for `xpauth`/`*.xpr`/`licen` **filenames** in the latest wheel — 9.9.x no longer ships one, but older wheels do:

| wheel | licence release | expiry |
|---|---|---|
| 9.4.6 | 9.4 | 30-apr-2027 |
| 9.5.8 | 9.5 | 30-jun-2027 |
| 9.6.3 | 9.6 | 31-aug-2027 |
| 9.7.2 | 9.7 | 31-aug-2027 |
| **9.8.1** | **9.8** | **28-feb-2027** |

Note 9.5.4's licence expires **28-feb-2026** — exactly the date in the error from `XPAUTH_XPR`. So the secret almost certainly came from this same source and simply went stale.

### Getting it working

The licence caps the solver release it authorises, which took two iterations to pin down:

1. 9.4.6 licence → `licensing error 21: Your license only authorizes up to release 9.4`
2. 9.8.1 licence → `...up to release 9.8`

The cap moved with the licence, proving it was read correctly — the mismatch was `Xpress_jll`. **`Xpress_jll` is versioned after the JLL build, not the solver**: 9.4.3 ships FICO release 9.9 (`get_version()` → 43.1.4). So the JLL is now pinned to `=9.8.0` (release 9.8, `get_version()` → 46.1.1) to match the newest licence available.

**Result: Xpress now actually runs in CI** — `[ Info: Running Xpress in Tests`, no skip, no licensing error, no secret required.

### What's left: 22 genuine test failures

```
2165 passed, 22 failed, 0 errored
```

Xpress adds ~640 tests, and 22 of them get **wrong answers** rather than erroring:

- **Princeton Handbook Quadratic** — 16 failures
- **Simple BLP JuMP** — 6 failures

e.g. [test/problems/jump.jl:556](../blob/master/test/problems/jump.jl#L556):

```
Expression: ≈(objective_value(model), -12, atol = atol)
 Evaluated: -21.0 ≈ -12 (atol=1.0e-6)
Expression: ≈(value(x), 4, atol = atol)
 Evaluated: 3.0 ≈ 4 (atol=1.0e-6)
```

These are `SOS1Mode` / `FortunyAmatMcCarlMode` bilevel reformulations whose expected values were recorded against Xpress ~8.11 (the old `SECRET_XPRS_WIN_8110`). Returning `-21` where `-12` is expected suggests the big-M bounds or SOS1 handling behave differently on release 9.8 — plausibly a genuinely different-but-valid local solution, or big-Ms that are now too loose.

I have stopped here rather than guessing: adjusting expected values or big-M/tolerance settings is a modelling judgement about what Xpress *should* return, and I would rather not silently loosen assertions that might be catching something real.

Options, roughly in order of my preference:

1. Review the 22 expectations against release 9.8 and update them where the new answer is legitimate.
2. Keep Xpress out of `solvers`/`solvers_quad` for these two problem sets, so it still exercises the rest.
3. Pin an older `Xpress_jll` closer to 8.11 — but no bundled community licence goes back that far, so this needs the secret renewed.

Say which you prefer and I will implement it. Everything up to this point is pushed and the branch is green apart from these 22.

joaquimg:
## Reproduced locally, and Xpress's log gives the answer

Two findings, and the first corrects something I said earlier.

### 1. Getting Xpress to actually run needed one more fix

Locally it kept failing with `XpressError(16): Failed to create XpressProblem`. With the log visible, the real message appears:

```
ERROR - call to XPRS function when global environment not initialised
        - please call XPRSinit() before attempting to call Xpress-Optimizer functions.
```

`XPRSinit()` was failing **silently**. Xpress.jl's *automatic* initialization does not pick up the licence when the library comes from `Xpress_jll`. Setting `XPRESS_JL_NO_AUTO_INIT` and calling `Xpress.initialize(; xpauth_path = ...)` explicitly — as [Xpress.jl's own test suite](https://github.com/jump-dev/Xpress.jl/blob/master/test/runtests.jl) does — fixes it:

```
[ Info: Xpress: Found license file .../xpauth.xpr
[ Info: Xpress: Development license detected.
INIT_OK  version=46.1.1   →   solves correctly
```

So I had earlier blamed my local `XpressError(16)` on the system Xpress installs on this machine. That was wrong — `/c/xpressmp` does not even exist, the PATH entries are stale. It was this init bug all along, and it was reproducible everywhere. Pushed as a commit.

### 2. The failures are Xpress's SOS1 handling, and the answers are genuinely wrong

With Xpress really running I reproduce them locally: **1537 passed, 18 failed** (CI shows 22; Linux also runs SCIP).

I also need to correct my earlier claim that these were "SOS1/big-M tests". The failing testsets — *Princeton Handbook Quadratic*, *Simple BLP JuMP* — iterate over `solvers_sos`, and Xpress joins that list with `SOS1Mode()`. **Every failure routes through SOS1Mode.** Same model, same Xpress, three modes:

| mode | result | correct? |
|---|---|---|
| `ProductMode` | x=4, y=4, obj=-12 | ✅ |
| `FortunyAmatMcCarlMode` | x=4, y=4, obj=-12 | ✅ |
| **`SOS1Mode`** | **x=3, y=6, obj=-21** | ❌ |

**These are not alternate optima.** For `_jump_06` the lower level is `min y`. At the returned `x=3`, the lower-level feasible range is `y ∈ [2.5, 6]`, so `y* = 2.5` — Xpress returned `y = 6`, the *worst* point for the follower and the best for the leader. Both points are jointly feasible, but `(3,6)` violates bilevel optimality outright.

Xpress's log says why:

```
Original problem has:  12 rows  12 cols  29 elements  5 entities
STOPPING - MIPRELSTOP target reached (MIPRELSTOP=0.0001  gap=-1.69177e-16)
  Max primal violation      (abs/rel) : 3.553e-15 / 1.184e-15
  Max dual violation        (abs/rel) :       0.0 /       0.0
  Max integer violation     (abs    ) : 1.000e+00      <-- !!
```

It reports `OPTIMAL` while admitting a **full integer violation of 1.0**. Primal and dual violations are ~1e-15, so the LP part is fine; it is the discrete part that is not enforced. `SOS1Mode` encodes complementarity as a real `MOI.SOS1([slack, dual])` constraint ([src/modes/sos1.jl:41](../blob/master/src/modes/sos1.jl#L41)), so an unenforced SOS1 lets both members be nonzero — complementarity breaks, and `y=6` becomes reachable. For contrast, `FortunyAmatMcCarlMode` on the same model reports **no** integer violation and gets the right answer.

Presolve is also aggressive here (12→3 rows, 12→8 cols), which is a common trigger for SOS mishandling.

### Recommendation

This looks like an Xpress 9.8 SOS1 bug, not a stale test expectation — so **do not update the expected values**; they are correct and Xpress's answer is not. Options:

1. **Report upstream** to Xpress.jl / FICO with this reproducer — the "OPTIMAL with integer violation 1.0" log is strong evidence.
2. Try `XPRS_MIPTOL` / disabling presolve (`XPRS_PRESOLVE=0`) to see whether it is presolve-specific — I can test this if useful.
3. Have Xpress join `solvers_fa` / `solvers_quad` but **not** `solvers_sos` until the SOS1 issue is resolved, so it still exercises the rest of the suite.

I have not implemented any of these — picking between "report upstream and leave red" and "drop Xpress from SOS1 tests" is your call. Everything else is pushed and green.

